PR #33, opened 2015-01-03
remove hardcoded config file paths for server, proxy and agent

f0:
removed the hardcoded config paths for agent, server and config

the new upstream default is /etc, and the epel rpm does create symliks forom /etc/zabbix/* to /etc, this causes errors. 

btw: sucessfull testet  the module on centos7 (with this changes)


f0:
hm sorry   the merge request i  against an old version, i create a new one

PR #35, opened 2015-01-04
ZABBIX proxy and ZABBIX server service names are now customizable / Allow changing the path to the database schema files

f0:
In newer EPEL rpms, the services are called zabbix-proxy-pgsql ..., so enable the service zabbix-proxy fails)
Allow changing the path to the database schema files  in newer EPEL rpms the location is /usr/share/zabbix-prostgresql for example


dj-wasabi:
Hi f0,

There is an params.pp in the ROOT? I don't think it belongs there.. :-)

If you would delete it and fix these:
manifests/database/mysql.pp - ERROR: trailing whitespace found on line 27
manifests/database/mysql.pp - ERROR: trailing whitespace found on line 28
manifests/database/postgresql.pp - ERROR: trailing whitespace found on line 27
manifests/database/postgresql.pp - ERROR: two-space soft tabs not used on line 34
manifests/proxy.pp - ERROR: trailing whitespace found on line 409

Then I can approve it.
Thanks!

Kind regards,
Werner

f0:
sorry, fixed

dj-wasabi:
Thanks! :-)

PR #36, opened 2015-01-06
allow custom  owner and group for zabbix server config,

f0:
this is needed because the EPEL Packages using zabbixsrv as config owner


dj-wasabi:
Thanks!

PR #38, opened 2015-01-10
setting manage_repo to false breaks server install

genebean:
You have this chunk of code in `zabbix::web`

```puppet
# Check if manage_repo is true.
  if $manage_repo {
    if ! defined(Class['zabbix::repo']) {
      class { 'zabbix::repo':
        zabbix_version => $zabbix_version,
      }
    }
  } # END if $manage_repo
```
but nothing that ensures it runs before the this chunk in the same file:

```puppet
default : {
      package { "zabbix-web-${db}":
        ensure  => $zabbix_package_state,
        before  => [
          File['/etc/zabbix/web/zabbix.conf.php'],
          Package['zabbix-web']
        ],
        require => Class['zabbix::repo'],
      }
```

This causes the following error: `Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid relationship: Package[zabbix-web-mysql] { require => Class[Zabbix::Repo] }, because Class[Zabbix::Repo] doesn't seem to be in the catalog`

I see how to fix this and will be submitting a pull request shortly but wanted to go ahead and list it as a bug in case anyone else is effected by this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ cache: bundler
 rvm:
   - 2.0.0
 
-script: "bundle exec rake rspec"
+script: "bundle exec rake spec"
 
 env:
   matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: ruby
 cache: bundler
 
 rvm:
-  - 1.8.7
-  - 1.9.3
   - 2.0.0
 
 script: "bundle exec rake rspec"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: ruby
+cache: bundler
+
+rvm:
+  - 1.8.7
+  - 1.9.3
+  - 2.0.0
+
+script: "bundle exec rake rspec"
+
+env:
+  matrix:
+    - PUPPET_GEM_VERSION="~> 2.7.0"
+    - PUPPET_GEM_VERSION="~> 3.7.3"
+
+notifications:
+  email:
+    on_failure: change
+    on_success: never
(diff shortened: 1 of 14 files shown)


genebean:
When viewing the changes to `zabbix::repo` it is much easier if you disable showing changes in whitespace.

dj-wasabi:
Thanks! 
Before I approve, did you also look at the rspec tests? They fail on my machine with this pull request. If you don't have experience with it, please tell me than I'll fix it (No worries ;-)). But I need to fix it first before approving this request.

Kind regards,
Werner

genebean:
I've done a little rspec before, what is your test environment? AKA, version of puppet, version of ruby, etc? is `bundle exec rake rspec`  your test command? I added a Travis-CI config file to help with testing to the branch I am working on. Is this what you see test wise too? https://travis-ci.org/genebean/puppet-zabbix

genebean:
it seems you call a file named `spec/spec.opts` in your rspec but that is not in the repo. I am going to manually make one for now just so the tests will run...

genebean:
So, I have gathered that I should call `bundle exec rake spec` to utilize the fixtures but seem to be missing something. Can you share what your testing setup is and, if you have one, your Vagrantfile?

Also, once this is setup the way it needs to be, I'll go back and squash my commits so that only useful ones show up as opposed to having multiple ones where I am hacking on Travis's setup and the like.

dj-wasabi:
Hi Genebean,

At the moment, I'm running the 'rake spec' command locally on my Mac, as I'm developing on the mac. I'am however busy with my internal Jenkins instance to run the 'rake spec' automatically with every push I make. It takes some time though. I was testing with puppet 3.4.3 and ruby 2.0.0 and just upgraded to puppet 3.7.3 on my mac. (And I just noticed with the new puppet I have an error.. :-))

Kind regards,
Werner

genebean:
Thanks. I have a Mac too so I will try it there and see what happens.

genebean:
In the process of looking at this I found in some of my tests that the zabbix-server process is trying to start too soon. I think this can be resolved via commit genebean/puppet-zabbix@1b89db8

As for the tests, I haven't had a chance to try on my Mac but any help you could give would be greatly appreciated.

f0:
+1

dj-wasabi:
Thanks! Will take a look at the rspecs..  no worries ;-)

genebean:
Awesome

dj-wasabi:
Updated the rspec tests, work now. :-)

PR #39, opened 2015-01-13
Fix name of agent config file in params.pp

mmerfort:
The default config path in params.pp was missing a “d” after “agent”.


dj-wasabi:
Thanks!

PR #40, opened 2015-01-15
This resolves dj-wasabi/puppet-zabbix#37.

genebean:
API docs say that ip must not be null.
https://www.zabbix.com/documentation/2.4/manual/api/reference/hostinterface/object#host_interface


dj-wasabi:
Thanks!

PR #48, opened 2015-02-10
Correctly rewrite the root when using SSL, fixes #47

### LINKED ISSUE #47: When using SSL, the root is not rewritten to SSL

slyoldfox:
This works for me more correctly, otherwise accessing the / keeps you on http:// and when POST'ing you get a weird 2nd login screen.


dj-wasabi:
Thanks!

Issue #50, opened 2015-02-13
Bad syntax in manifests/proxy.pp


fredprod:
In forge.puppetlabs repository
casesize instead of cachesize
check comments we have to use this param in MB not in B
for example, just pust 32 if you want 32M in CacheSize value

dj-wasabi:
Hi Fredprod,

I accepted your pull request (Thanks again!) and updated the comment in manifests/proxy.pp (and configuration template) to use MB.

Kind regards,
Werner

PR #51, opened 2015-02-13
line 350 modify casesize to cachesize #50

fredprod:
Syntax problem in proxy.pp for CacheSize and in templates


dj-wasabi:
Thanks!

PR #52, opened 2015-02-15
Fix agent listenip

JvdW:
Fix agent listenip construct to be the same as in proxy.pp. This fixes a bug (== vs != or defined vs undefined) and style.


dj-wasabi:
Thanks!
I fixed it locally but not commited it yet, so I'll use your version.

PR #53, opened 2015-02-16
Jvd w fix defined

JvdW:
Sorry about the previous change, that didn't work. I looked at the 'if $var == defined' construct but that isn't a valid puppet construct but I didn't know that. Puppet only knows about 'undef' of empty string ''.


dj-wasabi:
No worries! Thanks!

PR #56, opened 2015-02-20
Fix in params.pp with default parameter of zabbix proxy for ubuntu

fredprod:
Some mistakes in params.pp for zabbix proxy (mysql)


dj-wasabi:
Thanks!

PR #63, opened 2015-03-19
Fix manage_repo parameter on the zabbix class

roidelapluie:
Without this commit, the manage_repo in the main class is not doing
anything.

This commit fixes that by explicitely passing the parameter to the
zabbix::repo class.

See also GH-64 for a different way of solving this


dj-wasabi:
Hi roidelapluie,

Thanks! I like this version. :-)

Kind regards,
Werner

PR #64, opened 2015-03-19
Remove manage_repo parameter on the zabbix class

roidelapluie:
The manage_repo in the main class is not doing anything.

This commit fixes that by removing the parameter from the zabbix
class.

See also GH-63 for a different way of solving this.


dj-wasabi:
Hi roidelapluie,

I approved the other version, so I'll close this pull request.
Thanks!

Kind regards,
Werner

Issue #61, opened 2015-03-17
Apt-key error in Ubuntu 14.04


HielkeJ:
In Ubuntu 14.04 we get this error every Puppet run:

Warning: /Apt_key[Add key: 79EA5ED4 from Apt::Source zabbix]: The id should be a full fingerprint (40 characters), see README.

The fingerprint should be:
Key fingerprint = FBAB D5FB 2025 5ECA B22E  E194 D13D 58E4 79EA 5ED4

dj-wasabi:
Hi HielkeJ,

Thank you for making time to create this issue!
Will take a look at it.

Kind regards,
Werner

dj-wasabi:
Hi,

The user @exptom just created an Pull request for this issue and I just accepted this request.
So issue should be solved now.

Kind regards,
Werner

PR #66, opened 2015-03-20
Update apt key to full 40characters

exptom:
Latest version of puppetlabs/apt module shows warning on every puppet run (Warning: /Apt_key[Add key: 79EA5ED4 from Apt::Source zabbix]: The id should be a full fingerprint (40 characters), see README.) if using a short key so this update includes the full 40 character key.

See Issue #61


dj-wasabi:
Thanks!

PR #67, opened 2015-03-29
bugfix for vhosts in apache 2.4

ju5t:
Currently it's not possible to manage vhosts in Apache 2.4. The module is using deprecated functions that only work for 2.2. This commit solves that and keeps it backwards compatible.


dj-wasabi:
Thanks!

Issue #70, opened 2015-03-31
Multi-node setup: manage_resources invalid on Class[Zabbix::Server]


elricsfate:
Documentation reports that "manage_resources" should be available to the following 

zabbix::server, zabbix::proxy & zabbix::agent

Using manage_resources on zabbix::server returns an error

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid parameter manage_resources on Class[Zabbix::Server] at /etc/puppetlabs/puppet/environments/production/manifests/zabbix.pp:41 on node HOSTNAME
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

Puppet code used

```
 class { 'zabbix::server':
    database_host  => 'HOSTNAME',
    database_type  => 'mysql',
    manage_resources => true,
  }
```

dj-wasabi:
Hi elricsfate,

Thank you for making time to create this issue. Also many thanks for creating an pull request!

Kind regards,
Werner

elricsfate:
Howdy Werner,

No problem. Thanks for the swift inclusion. 

Sincerely, 

elricsfate

PR #71, opened 2015-03-31
Fixed small error regarding manage_resources and it's usage with classes

elricsfate:
See Issue #70


dj-wasabi:
You are absolutely right, my bad. Thanks!

PR #79, opened 2015-04-16
Allow to not purge include dir. 

altvnk:
For example if userparameter configs is managed via VCS repo. Default value is set to "true", as was in original implementation.


dj-wasabi:
Thanks!

PR #82, opened 2015-04-20
Fixes #80 setting Hostname and HostnameItem causes a warning on agentd s...

### LINKED ISSUE #80: setting Hostname and HostnameItem causes a warning on agentd start
Hostname and HostnameItem are both set by default, this leads into this warning:

Shutting down Zabbix agent:                                [  OK  ]
Starting Zabbix agent: zabbix_agentd [3912]: Warning: both Hostname and HostnameItem defined, using [xxx]

HostnameItem should not set by default, and Hostname should be disabled when HostnameItem is set

f0:
...tart


dj-wasabi:
Thanks!

PR #87, opened 2015-05-13
Added zabbix_hostgroup

hkumarmk:
This patch contains:

* Create a provider for zabbix_hostgroup
* Moved common code from type to util/zabbix.rb, so that type code is simpler


dj-wasabi:
Thanks!

PR #88, opened 2015-05-18
Pass zabbix_version and manage_repo to zabbix::repo

mmerfort:
Currently the zabbix::repo class only uses the defaults from zabbix::params and does not use the value from zabbix::server. Manage_repo is set to false by default and it's not possible to change this.

This pull request fixes this problem by passing manage_repo and zabbix_version to zabbix::version.


dj-wasabi:
Hi @mmerfort 

If I read this correctly, this looks almost the same as issue #93 ?
With the #93 I added the 'inherit from zabbix::params ' into the zabbix::repo manifest file.

Does this solve your problem?
If I accept this pull request, I think I'll receive a lot of issues about 'class zabbix::repo is already loaded'.

Kind regards,
Werner

mmerfort:
Hi @dj-wasabi 
Sorry for taking so long. I just saw that I overlooked some important things in my pull request (passing the $manage_repo and $zabbix_version parameters to the server class).
Maybe I'm understanding something wrong but doesn't init.pp just do nothing with the parameter? If I pass manage_repo => false to the zabbix class (init.pp) it will only set this value for this class alone and not for repo.pp.
I think a good solution to this problem would be to set the defaults in repo.pp to the values of the variables in init.pp ($::zabbix::manage_repo).

dj-wasabi:
Thanks! It seems that without this, it always uses the default (2.4) release to install..

rnelson0:
@dj-wasabi This does indeed cause our rspec-puppet tests to all fail with duplicate declarations. Thankfully we caught this before pushing to production. I'm not sure what the suggested workaround is for this.

I think this could also cause permanent breakage for anyone declaring both zabbix::server and zabbix::repo. Is it possible for this to be reviewed and another more idempotent solution implemented?

mmerfort:
I have added another pullrequest which should fix this issue: https://github.com/dj-wasabi/puppet-zabbix/pull/110

PR #95, opened 2015-06-02
import templates and create hostgroup if missing

1n:
Implemented zabbix templates import.
Import is idempotent, template being imported is compared with template with the same name exported from zabbix server. At this point there is no template destroy option.
Also pull request include option to create hostgroup if it is missing, which defaults to true in params.

### DIFF
--- a/lib/puppet/provider/zabbix.rb
+++ b/lib/puppet/provider/zabbix.rb
@@ -61,6 +61,32 @@ class Puppet::Provider::Zabbix < Puppet::Provider
         template_array.include?("#{template_id}")
     end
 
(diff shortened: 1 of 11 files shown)


dj-wasabi:
Thanks! :+1:

elricsfate:
@dj-wasabi 

The feature @1n implemented is for importing Zabbix templates right? If so it might make sense to integrate this portion with my recently pulled Zapache support (currently requires user to import template manually).

~ef

dj-wasabi:
Hi,

Yes, you are right. (If I'm reading it correctly.)
This can save a people a lot of time!

When I have some time left, I try to add something into the readme.

Kind regards,
Werner

PR #96, opened 2015-06-02
Added support to Amazon Linux with epel 6.

Wprosdocimo:
I was changed this to work with my AWS instances, I think that you can add this change in your code.


dj-wasabi:
Thanks! :+1:

PR #98, opened 2015-06-05
merge of hiera hashes from entire hierarchy

szemlyanoy:
this should fix the lookup of hash only from the most specific level


dj-wasabi:
Thanks!

PR #99, opened 2015-06-09
Custom apache IP and port

mschuett:
In order to use a reverse proxy I modified the module with a configurable Apache listen IP and port number. Might be useful...


dj-wasabi:
Thanks!

PR #104, opened 2015-07-09
Add "eno*" to interface name matching

sgnl05:
This will fix issue with RHEL7 servers having "eno167800..." names for interfaces


dj-wasabi:
Thanks!

PR #110, opened 2015-07-21
Pass manage_repo and zabbix_repo to repo.pp and prevent double include

mmerfort:
All classes which call zabbix::repo should pass manage_repo and zabbix_version to zabbix::repo.
To prevent double includes the classes will check if zabbix::repo has already been included.
Instead of checking if the zabbix::repo class should be included (if $manage_repo) the class will be included either way and zabbix::repo will decide if it manages the repositories.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -251,10 +251,15 @@ class zabbix::agent (
     }
   }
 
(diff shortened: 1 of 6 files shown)


dj-wasabi:
Thanks!

PR #132, opened 2015-11-15
tap0 or tun0 (OpenVPN interfaces) interface as listenip

z3rogate:
Make it possible to set an tap0 or tun0 (OpenVPN interfaces) interface as listenip.


dj-wasabi:
Thanks!

PR #133, opened 2015-11-15
Listen on all IPs

z3rogate:
Listen on all IPs if nothing specified ip or interface is set in the Agent Class.


dj-wasabi:
Thanks!

PR #137, opened 2015-11-25
enable apache_php_max_input_vars

bastelfreak:
This adds the param apache_php_max_input_vars to the zabbix::web class.
It will allow us to configure the amount of max input vars for GET/POST
requests. On large installations you need 2000 or more vars on mass
updates. 1000 is the default value on CentOS which gets set in the
zabbix::params.pp


dj-wasabi:
Thanks!

PR #141, opened 2015-12-08
fixed SSL server template options for 2.2

IceBear2k:
SSLCertLocation and SSLKeyLocation were introduced with 2.4 and aren't compatible with 2.2, resulting in the server refusing to start due to unknown options.

https://www.zabbix.com/documentation/2.4/manual/introduction/whatsnew240#ssl_verification_and_authentication_options


dj-wasabi:
Thanks!

PR #146, opened 2016-01-20
allow serveractive to be optional

ericsysmin:
server active gives errors if used on agents that aren't using active agent calls, this prevents active check configuration update from [127.0.0.1:10051] started to fail (cannot connect to [[127.0.0.1]:10051]: [111] Connection refused) from occurring


dj-wasabi:
Thanks!

PR #147, opened 2016-01-20
Allow agent_serveractive value to be blank

ericsysmin:
to prevent  10832:20160120:225723.336 active check configuration update from [127.0.0.1:10051] started to fail (cannot connect to [[127.0.0.1]:10051]: [111] Connection refused) from occurring on agents


dj-wasabi:
Thanks!

PR #154, opened 2016-01-29
Added zabbix_template_host type

hkumarmk:
This new type can be used to link/unlink template to the host. zabbix_host
provide a way to assign list of templates while creating the host. But in
a scenario where:
* One need to add zabbix_host from agent setup code
* Link appropriate templates from the service setup code. For example, mysql server
deployment code can use below code to link the template "Template App MySQL" to the host.

```
zabbix_template_host{"Template App MySQL@${::hostname}": }

```


dj-wasabi:
Wow, thanks! :+1:

hkumarmk:
You're welcome :+1:  thanks for merging @dj-wasabi

Issue #153, opened 2016-01-29
Database is always managed


subkowlex:
Hi,
I use multiple hosts setup where zabbix proxy db is on an external server.
That's why I use zabbix::proxy::manage_database: false

But on every run puppet tries to update:

`Error: cd /usr/share/zabbix-*-mysql && if [ -f schema.sql.gz ]; then gunzip schema.sql.gz ; fi && mysql -h 'db-host' -u 'db_user' -p'password -D 'db' < schema.sql && touch /etc/zabbix/.schema.done returned 1 instead of one of [0]
Error: /Stage[main]/Zabbix::Database::Mysql/Exec[zabbix_proxy_create.sql]/returns: change from notrun to 0 failed: cd /usr/share/zabbix-*-mysql && if [ -f schema.sql.gz ]; then gunzip schema.sql.gz ; fi && mysql -h 'db_host' -u 'db_user' -p'db_pass' -D 'db' < schema.sql && touch /etc/zabbix/.schema.done returned 1 instead of one of [0]`

Any idea what's wrong?

dj-wasabi:
Hi @subkowlex 

Thank you for making time to create this issue.
When reviewing the code, it seems that when manage_database is set to false in the proxy.pp, it won't create the database but it will execute the SQL scripts. That is not how it should be, so I will fix this.

I'll let you know when this is fixed.
Kind regards,
Werner

dj-wasabi:
Hi,

This is fixed in 2.0.0 which is available since yesterday.

PR #159, opened 2016-02-01
Update proxy.pp, fix Error: ...install zabbix-proxy- ..

subkowlex:
Hi,
thank you for the quick fixing of the issue  Database is always managed #153 
I tried running puppet fix the fix but got:

`Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install zabbix-proxy-' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package zabbix-proxy
Error: /Stage[main]/Zabbix::Proxy/Package[zabbix-proxy-]/ensure: change from purged to present failed: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install zabbix-proxy-' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package zabbix-proxy`

After a quick look into the fix, I saw that there part where db variable is set was not reachable if the database is managed.

So I extracted the file name building depending on DB type to an extra loop.

### DIFF
--- a/manifests/proxy.pp
+++ b/manifests/proxy.pp
@@ -409,10 +409,21 @@ class zabbix::proxy (
 
   # Get the correct database_type. We need this for installing the
   # correct package and loading the sql files.
+  case $database_type {
+    'postgresql': {
+      $db = 'pgsql'
+    }
+    'mysql': {
+      $db = 'mysql'
+    }
+    'sqlite': {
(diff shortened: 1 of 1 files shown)


dj-wasabi:
Thanks!

PR #162, opened 2016-02-02
Travis CI setup: ensure all rspec tests pass

### LINKED ISSUE #160: An IP address is required for a zabbix host
In #133, the default `$listenip` value was removed. When this is not present, the following error is observed:
````
Error: Could not set 'present' on ensure: Server answer API error:
 {
  "code": -32602,
  "message": "Invalid params.",
  "data": "Incorrect arguments passed to function."
}
 on request:
 {
  "method": "host.create",
  "params": [
    {
      "host": "zabbix.nelson.va",
      "interfaces": [
        {
          "type": 1,
          "main": 1,
          "ip": null,
          "dns": "zabbix.nelson.va",
          "port": "10050",
          "useip": 1
        }
      ],
      "status": 0,
      "available": 1,
      "groups": [
        {
          "groupid": 2
        }
      ],
      "proxy_hostid": null,
      "templates": [
        10001,
        10102
      ]
    }
  ],
  "id": 8903,
  "jsonrpc": "2.0",
  "auth": "752a36e6c48997194ee8b7ecb813fc40"
} at 27:/etc/puppetlabs/code/environments/zabbix/modules/zabbix/manifests/resources/agent.pp
````
By replacing the `else{}` block, this functions properly. PR forthcoming.

rnelson0:
This is an update to the Travis CI configuration and a rewrite of all rspec tests that standardizes them using the same layout. It also includes the fixes to the code required to get the rspec tests to pass with Puppet 3 and 4.

It does remove Puppet 2.7 from the test matrix. 2.7 is no longer supported and rspec generates a ton of failures that cannot be fixed without massive rewrites to the code. I believe this is fine, but it is up to you to accept it or not.

Fixes #160

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,18 @@
 language: ruby
 sudo: false
 cache: bundler
-
-rvm:
-  - 2.0.0
-
-script: "bundle exec rake test"
-
-env:
-  matrix:
-    - PUPPET_GEM_VERSION="~> 2.7.0"
-    - PUPPET_GEM_VERSION="~> 3.0"
-    - PUPPET_GEM_VERSION="~> 4.0"
-
 notifications:
-  email:
-    on_failure: change
-    on_success: never
+  email: false
+branches:
+  only:
+    - master
+bundler_args: --without development system_tests
+before_install: rm Gemfile.lock || true
+rvm:
+  - 1.9.3
+  - 2.1.0
+script: "bundle exec rake test"
+env:
+  - PUPPET_GEM_VERSION="~> 3.0" STRICT_VARIABLES=yes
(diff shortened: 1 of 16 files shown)


rnelson0:
Working on making this pass Travis CI. Seeing lint problems I did not observe locally now.

rnelson0:
@dj-wasabi This is ready for review now. I've used it to deploy my simple AIO zabbix server, so I know that the rspec test changes aren't affecting single node installations at least, and of course rspec tests themselves are passing. I did a little reformating to make the tests run faster but there is no net change in number of tests, just how they're grouped.

rnelson0:
With this and #163, a new version should be cut. As described above there should be no effective changes but it may be safer to go to `2.1.0` instead of `2.0.1` just in case...

dj-wasabi:
Many thanks for this! :+1:

PR #163, opened 2016-02-02
Add Puppet Forge Version and Downloads badges

rnelson0:


rnelson0:
@dj-wasabi This is a readme only change, separate from the rspec branch, so I expect Travis CI to fail - I wasn't smart enough to add [skip ci] to it first :)

dj-wasabi:
Again, thank you for making this possible!

PR #167, opened 2016-02-16
Adjust server config and databases sqls for 3.0

slashr00t:
I've prepared a fix for version 3.0. It looks like guys from Zabbix have changed package structure with sqls - now there's only a single file in an upper directory.
I also needed to adjust server config template. Maybe it's not the most elegant solution, but it works.

### DIFF
--- a/manifests/database/mysql.pp
+++ b/manifests/database/mysql.pp
@@ -29,7 +29,7 @@ class zabbix::database::mysql (
   if ($database_schema_path == false) or ($database_schema_path == '') {
     case $::operatingsystem {
       'CentOS','RedHat','OracleLinux' : {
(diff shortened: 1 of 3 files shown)


slashr00t:
I guess it's a big topic, most of the tests fail.

dj-wasabi:
Hi @slashr00t ,

Thank you for making time to create this PR, but I can't accept PR that fails the CI tests.
I'll leave this PR open if you want to make it work.

slashr00t:
Hi @dj-wasabi ,

I've fixed the code and it passed all the tests. I've also added a simple comment regarding this sqls placement difference.
If you have more suggestions please let me know.

dj-wasabi:
Thanks!

PR #171, opened 2016-02-17
Patch 6

ericsysmin:


ericsysmin:
As of this comment templates now work for zabbix-server running 3.0, as well as manage_database = false.

It also looks like it supports your testing with 2.4

dj-wasabi:
Many thanks again! :+1:

PR #188, opened 2016-03-08
updated to include Amazon Linux

ericsysmin:
will run on EL 6 library as EL 7 isn't supported by Amazon Linux and Amazon Linux doesn't have systemd


dj-wasabi:
Thanks!

Issue #194, opened 2016-03-21
Add support for zabbix sender


vide:
Please add support for zabbix sender since it usually comes in a different package (both under Debian and RedHat)

dj-wasabi:
As PR is merged, I close this issue.

PR #195, opened 2016-03-21
Add Zabbix sender support

vide:
This implement #194 and provide a simple zabbix::sender class that
basically installs the required package. Plus, I've updated doc & tests.


dj-wasabi:
Thanks!

PR #202, opened 2016-04-07
updated template

ericsysmin:


dj-wasabi:
Hi @ericsysmin 

Is this PR ready to be merged? Or do you have some other changed rolled up on your sleeves ;-)

ericsysmin:
@dj-wasabi I need to check a few more items before I merge in, there's some stuff I found that may not be in the other templates.

bastelfreak:
Hi @ericsysmin,

How is the state of the PR? Why did you downgrade the default zabbix version?

jyaworski:
@bastelfreak we should consider turning on auto-squash for these kinds of PRs.

bastelfreak:
squashing this one is fine. But I want to understand the reason for the downgrade of zabbix first.

ericsysmin:
When you upgrade it caused some problems. Hence why I wanted to change it back. But at this point it's probably best to close.

PR #207, opened 2016-04-18
add Arch Linux to metadata.json

bastelfreak:
Arch is a rolling release distribution, the operatingsystemrelease fact
only returns the current booted kernel. Relying on that isn't useful
here.


bastelfreak:
I added Archlinux support a few weeks back but forgot to adjust the metadata.json.

dj-wasabi:
Thanks!

PR #208, opened 2016-04-22
userparameters not working.

cbergmann:
This typo prevents working userparameters.


bastelfreak:
How could this pass CI with STRICT_VARS enabled?

cbergmann:
I don't know CI well but The error was in a String. I don't know if CI checks this deep.

bastelfreak:
Thanks for the fix @cbergmann

PR #210, opened 2016-04-22
enable STRICT_VAR for spec tests

bastelfreak:
this may allow CI to detect the issue in https://github.com/dj-wasabi/puppet-zabbix/pull/208

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -19,4 +19,7 @@ end
 
 RSpec.configure do |c|
   c.hiera_config = File.expand_path(File.join(__FILE__, '../fixtures/hiera.yaml'))
+  c.before :each do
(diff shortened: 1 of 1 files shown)


bastelfreak:
okay, no difference. I've no clue:
* if this is needed
* why CI didn't catch the issue in https://github.com/dj-wasabi/puppet-zabbix/pull/208

I copied that change here from another module. maybe puppet checks for the env var by itself, or any fo the rake tasks does it.

dj-wasabi:
Hi @bastelfreak 

What do you want to do with this PR. Keep it open until you find something? Or is it readto to merge or close?

Keep up the good work! 👍 👍

bastelfreak:
I digged through the gems, they should handle STRICT_VAR env by default. this doesn't seem to be needed. but I still don't know why travis didn't catch the error in #208.

PR #211, opened 2016-04-27
fix typo in zabbix_host provider

damoxc:
I am assuming here that `is_af?` should be `is_a?`.


bastelfreak:
ups. thanks for fixing that!

Issue #216, opened 2016-05-12
 undefined local variable or method `int_name'


lkthomas:
2016-05-12T09:32:43.637658+08:00 openvpn puppet-agent[9645]: Starting Puppet client version 4.4.2
2016-05-12T09:32:47.559683+08:00 openvpn puppet-agent[9649]: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Function Call, Failed to parse inline template: undefined local variable or method `int_name' for #<Puppet::Parser::TemplateWrapper:0x30ffee23> at /etc/puppetlabs/code/environments/production/modules/zabbix/manifests/agent.pp:269:20 on node <MASKED>.com

any idea what is this?

lkthomas:
It's complaining this line from agent.pp:

$listen_ip = inline_template('<%= scope.lookupvar(int_name) %>')

bastelfreak:
@lkthomas I maybe fixed it in https://github.com/voxpupuli/puppet-zabbix/pull/217. Do you want to test?

Interesting that our tests didn't complain.

lkthomas:
Can you merge it to master? I don't know how to get a particular code base.

Issue #212, opened 2016-05-02
ListenIP to accept $::fqdn as a value


pvl7:
Hello

I have multitenant systems where Zabbix agent binds to the wrong interfaces. Using '*' isn't a solution because, for example, on DMZ systems it may create a security hole. Can't rely on interface naming either as it is not consistent across systems and might not pick up the right nic with default module parameters. 

The easy solution that works is to pass the $::fqdn or other fully qualified name to the listenip parameter for agent but the Puppet module must translate it to the IP address as Zabbix agent doesn't accept fdqn as a value. I've done that change to my local fork and it works perfectly. Wondering if you ok to merge it to the original repo? :)

```
[11:14]:git diff manifests/agent.pp
diff --git a/manifests/agent.pp b/manifests/agent.pp
index 701dd15..b501021 100644
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -263,12 +263,19 @@ class zabbix::agent (
   # to network name. If more than 1 interfaces are available, we
   # can find the ipaddress of this specific interface if listenip
   # is set to for example "eth1" or "bond0.73".
+  #
+  # On multitenant systems sometimes useful to feed listenip
+  # with the host's FQDN record to bind on the proper interface.
+  #
   if ($listenip != undef) {
     if ($listenip =~ /^(eth|bond|lxc|eno|tap|tun).*/) {
       $int_name  = "ipaddress_${listenip}"
       $listen_ip = inline_template('<%= scope.lookupvar(int_name) %>')
     } elsif is_ip_address($listenip) or $listenip == '*' {
       $listen_ip = $listenip
+    } elsif is_domain_name($listenip) {
+      $domain_name = $listenip
+      $listen_ip = inline_template('<%= Resolv.getaddress(domain_name) %>')
     } else {
       $listen_ip = $::ipaddress
     }
```

Thanks,
Pavel

bastelfreak:
Hi,

are you interested to turn the block into a function including your changes? We discussed that in https://github.com/voxpupuli/puppet-zabbix/pull/217#discussion_r62984382

pvl7:
Hi

yes, I thought about it as there are too many inline templates in the block.

bastelfreak:
That would be cool. Please use our master branch as a base. I removed the inline templates already.

jyaworski:
[...]

PR #217, opened 2016-05-12
Fix216

bastelfreak:
can somebody review? This should fix https://github.com/voxpupuli/puppet-zabbix/issues/216

### DIFF
--- a/lib/puppet/provider/zabbix.rb
+++ b/lib/puppet/provider/zabbix.rb
@@ -1,4 +1,3 @@
-# encoding: utf-8
 # zabbix provider type for puppet
 class Puppet::Provider::Zabbix < Puppet::Provider
   # Require the zabbixapi gem
--- a/lib/puppet/provider/zabbix.rb
+++ b/lib/puppet/provider/zabbix.rb
@@ -1,4 +1,5 @@
 # zabbix provider type for puppet
+# rubocop:disable Style/ClassAndModuleChildren
 class Puppet::Provider::Zabbix < Puppet::Provider
   # Require the zabbixapi gem
   def self.require_zabbix
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -265,8 +265,7 @@ class zabbix::agent (
   # is set to for example "eth1" or "bond0.73".
   if ($listenip != undef) {
     if ($listenip =~ /^(eth|bond|lxc|eno|tap|tun).*/) {
-      $int_name  = "ipaddress_${listenip}"
-      $listen_ip = inline_template('<%= scope.lookupvar(int_name) %>')
+      $int_name  = getvar("::ipaddress_${listenip}")
     } elsif is_ip_address($listenip) or $listenip == '*' {
       $listen_ip = $listenip
     } else {
--- a/manifests/proxy.pp
+++ b/manifests/proxy.pp
@@ -421,8 +421,7 @@ class zabbix::proxy (
   # is set to for example "eth1" or "bond0.73".
   if ($listenip != undef) {
     if ($listenip =~ /^(eth|bond|lxc|eno|tap|tun).*/) {
-      $int_name  = "ipaddress_${listenip}"
-      $listen_ip = inline_template('<%= scope.lookupvar(int_name) %>')
+      $int_name  = getvar("::ipaddress_${listenip}")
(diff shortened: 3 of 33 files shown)


manifests/agent.pp
```diff
@@ -265,8 +265,7 @@
   # is set to for example "eth1" or "bond0.73".
   if ($listenip != undef) {
```
igalic: this whole thing could be turned into a case statement to a little clearer

igalic: but lets leave that for another time maybe?

manifests/proxy.pp
```diff
@@ -421,8 +421,7 @@
   # is set to for example "eth1" or "bond0.73".
   if ($listenip != undef) {
     if ($listenip =~ /^(eth|bond|lxc|eno|tap|tun).*/) {
-      $int_name  = "ipaddress_${listenip}"
-      $listen_ip = inline_template('<%= scope.lookupvar(int_name) %>')
+      $int_name  = getvar("::ipaddress_${listenip}")
```
igalic: heck, we could even put it into a function, seeing how we've got it twice :P

lib/puppet/provider/zabbix.rb
```diff
@@ -1,5 +1,5 @@
-# encoding: utf-8
 # zabbix provider type for puppet
+# rubocop:disable Style/ClassAndModuleChildren
```
igalic: what's the point of that style?

spec/classes/web_spec.rb
```diff
@@ -117,7 +116,8 @@ def package_provider_for_gems
           database_name: 'zabbix-server',
           database_user: 'zabbix-server',
           database_password: 'zabbix-server',
-          zabbix_server: 'localhost')
+          zabbix_server: 'localhost'
```
igalic: while you're at it, add a trailing comma

manifests/proxy.pp
bastelfreak: I mentioned that in https://github.com/voxpupuli/puppet-zabbix/issues/212. We should track that in a further PR.

spec/classes/web_spec.rb
bastelfreak: fixed in bd2918e42dd77915b873b44c05e0051ac5586230

lib/puppet/provider/zabbix.rb
bastelfreak: That would force us to use:
```ruby
class Puppet
  class Provider
    class Zabbix
    end
  end
end
```

But it would overwrite the Puppet/Provider classes.

manifests/agent.pp
bastelfreak: yep. referenced in another issue.

spec/classes/web_spec.rb
bastelfreak: @igalic: rubocop doesn't want the comma:
https://travis-ci.org/voxpupuli/puppet-zabbix/jobs/129740215

rubocop is here a bit stricter than in other modules. remove the cop or the commit? I vote for the commit.

spec/classes/web_spec.rb
```diff
@@ -117,7 +116,8 @@ def package_provider_for_gems
           database_name: 'zabbix-server',
           database_user: 'zabbix-server',
           database_password: 'zabbix-server',
-          zabbix_server: 'localhost')
+          zabbix_server: 'localhost',
```
igalic: sigh, of course that now causes a build failure ~_~

bastelfreak:
@igalic IMO I should revert the last commit, then we can merge it. Opinions?

jyaworski:
Sounds good to me.

spec/classes/web_spec.rb
rnelson0: Try
```
let :params do
  super().merge({
    # stuff
  })
```
Pass the hash!

bastelfreak: this will throw:
```ruby
spec/classes/database_mysql_spec.rb:11:21: C: Style/BracesAroundHashParameters: Redundant curly braces around a hash parameter.
        facts.merge({ ...
```

the rubocop config in this module is a bit stricter than our default one, and I want to keep that behavior. I will remove the commit as @jyaworski also seems to be fine with that.

PR #218, opened 2016-05-12
Introduce rspec-puppet-facts for some tests

bastelfreak:


bastelfreak:
this should get green except for:
* rubocop issues. will be fixed in https://github.com/voxpupuli/puppet-zabbix/pull/217
* 4 spec tests

can somebody review this please? (/me looks at @jyaworski) I've no clue why these for tests fail. Maybe sientifix linux has the wrong osfamily fact.

bastelfreak:
ha awesome... I did the local tests withour STRICT_VARIABLES=yes... is there a nice way to automate the lsb* fact mocking? I would like to get rid of the manual style. Also what is the correct way to deal with the is_pe fact?

manifests/repo.pp
```diff
@@ -93,7 +88,10 @@
           }
         } else {
           $operatingsystem = downcase($::operatingsystem)
-
+          case $::operatingsystemrelease {
+            /\/sid$/ : { $releasename = regsubst($::operatingsystemrelease, '/sid$', '') }
+            default  : { $releasename = $::lsbdistcodename }
```
jyaworski: Is lsbdistcodename a thing on newer facter? I seem to recall the lsb* facts being deprecated.

bastelfreak: Ionly know that it requires the lsb-release package installed and in general the facts are only present on Debian based systems. I was looking for alternative core facts but didn't find any :(

Issue #222, opened 2016-05-18
What is selinux_config_mode?


ITler:
Hy,

I've upgraded to 2.3.1-rc0 and tried to install zabbix from scratch in my vagrant box. This have had worked the day before, but today after upgrade I get

    Evaluation Error: Unknown variable: '::selinux_config_mode'. at /etc/puppetlabs/code/environments/dev/modules/zabbix/manifests/web.pp:429:14

The specs do not cover the case, when variable is not defined.

I'm working on a PR to make module run.

bastelfreak:
ah shit, I implemented selinux support, but the facts are only available on RHEL I think. I will provide a patch soon.

ITler:
My PR is getting green in travis

bastelfreak:
ah nice, you did https://github.com/voxpupuli/puppet-zabbix/pull/223

bastelfreak:
this should be fixed now: https://github.com/voxpupuli/puppet-zabbix/pull/224

sgnl05:
@bastelfreak I don't think we need to check $::selinux_config_mode, as long as we check that $::osfamily is 'RedHat'. Setting selbooleans on RHEL systems with selinux disabled makes no difference. If selinux is in permissive mode, this seboolean would only affect what warnings selinux would log.

sgnl05:
In other words, I think zabbix_can_network should be set to 'on' no matter what state selinux is in. This of course goes for the proxy seboolean in #235 as well.

bastelfreak:
ah interesting, I didn't know that. This would add a useless resource to the catalog but takes complexity from the module (and mocking from the spec tests).

---

@igalic @jyaworski what do you think here, should we remove the check for selinux_config_mode and always set the selinux selboolean?

jyaworski:
Does Puppet handle selboolean and selmodule gracefully when selinux is disabled?

igalic:
@jyaworski my suspicion, with selstuff being rather built-in, would be yes.
but, let's just look at the code!!!
*****
okay, so basically, the source code says, this provider does nothing much other than call `getsebool`/`setsebool`… so it boils down to how those binaries handle it.
****
```
sudo -H getsebool -a
getsebool:  SELinux is disabled
#  => return status 1
```

PR #223, opened 2016-05-18
Make module run with ubuntu.

ITler:
This belongs to #222 and should make module run with ubuntu. I'd like to request you to extend your specs.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -507,7 +507,7 @@ class zabbix::server (
   }
 
   # check if selinux is active and allow zabbix
-  if $::selinux_config_mode == 'enforcing' {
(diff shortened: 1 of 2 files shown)


bastelfreak:
thanks for the fix! :+1: 
Could you also add a spec test for these two resources?

ITler:
Obviously, I'm not sure how to do this in an effective way... Also, I'm sitting in a project having time pressure. I just can't familiarize myself with adepting the specs at the moment.

ITler:
Can we leave ticket #222 open until specs are upgraded, but already merge this PR to master?

bastelfreak:
I will do the spec tests

PR #224, opened 2016-05-18
Introduce rspec-puppet-facts for some tests

bastelfreak:


spec/classes/web_spec.rb
```diff
@@ -14,154 +14,126 @@ def package_provider_for_gems
...
-
-    describe 'when manage_resources and is_pe are true' do
-      let :facts do
-        super().merge(
-          is_pe: true,
-          pe_version: '3.7.0'
+        facts.merge(
+          concat_basedir: '/tmp',
+          is_pe: false,
+          selinux_config_mode: 'disabled'
```
bastelfreak: I would like to mock this only on osfamily redhat, but I failed several times. does anybody have a good suggestion?

igalic: why not put those into spec_helper.rb?

bastelfreak: that is a good approach to prevent code duplication, we could store all needed facts in the spec helper and not in every spec test file. will do that.

bastelfreak: @igalic I pushed the changes, could you take a look again and maybe merge?

PR #225, opened 2016-05-19
Modulesync + Release

bastelfreak:
this makes the a modulesync with version 0.6.1 + a new release


bastelfreak:
for the release. there was a minor bump in the required stdlib version, does this require a minor bump on our side as well?

PR #229, opened 2016-05-27
Allows setting ListenIP as "lo" loopback interface.

felipe1982:
* listenip => 'eth0' was being ignored in zabbix-agentd.conf because $listen_ip was not being set when regex matched. now $listen_ip will be set to the IPaddress returned from $::ipaddress_eth0
* added option to use interface "lo" (127.0.0.1)

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -265,8 +265,7 @@ class zabbix::agent (
   # is set to for example "eth1" or "bond0.73".
   if ($listenip != undef) {
     if ($listenip =~ /^(eth|lo|bond|lxc|eno|tap|tun).*/) {
-      $int_name  = getvar("::ipaddress_${listenip}")
-      $listen_ip = $int_name
+      $listen_ip = getvar("::ipaddress_${listenip}")
     } elsif is_ip_address($listenip) or $listenip == '*' {
       $listen_ip = $listenip
     } else {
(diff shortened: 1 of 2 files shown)


manifests/agent.pp
```diff
@@ -264,8 +264,9 @@
   # can find the ipaddress of this specific interface if listenip
   # is set to for example "eth1" or "bond0.73".
   if ($listenip != undef) {
-    if ($listenip =~ /^(eth|bond|lxc|eno|tap|tun).*/) {
+    if ($listenip =~ /^(eth|lo|bond|lxc|eno|tap|tun).*/) {
       $int_name  = getvar("::ipaddress_${listenip}")
+      $listen_ip = $int_name
```
jyaworski: You could just set $listen_ip directly, here.

felipe1982: Are you suggesting not using $int_name, and replacing it with:
`$listen_ip = getvar("::ipaddress_${listenip}")` ? If $int_name is not used, then go for it. I didn't want to break the module by removing a variable I didn't understand.

jyaworski: Yes. It's not used anywhere else.

jyaworski:
Could you also make the same change here?

https://github.com/voxpupuli/puppet-zabbix/blob/f58bdb0a5a6094353b16757d8d81cf95d1c0a4f5/manifests/proxy.pp#L424

felipe1982:
@jyaworski  I hope I did it correctly :) I'm new to github. And I am only using Web UI.

PR #232, opened 2016-05-27
Fix Concat pinning

bastelfreak:


bastelfreak:
@jyaworski I'm not sure what is correct here. the general flow is to have git sources in fixtures to test against latest, but pin to something in metadata.json.

jyaworski:
This LGTM.

alexjfisher:
concat wasn't a dependency of *this* module.  It is needed in .fixtures.yml as it's a dependency of a dependency. See https://github.com/voxpupuli/puppet-zabbix/issues/658

PR #235, opened 2016-06-03
Set selboolean for zabbix proxy

sgnl05:
Just like Zabbix server, proxy needs selboolean zabbix_can_network set to 'on'.


bastelfreak:
Hi,
thanks for the fix, can you also add a spec test for the change?

sgnl05:
All done, I think :)

bastelfreak:
thanks!

Issue #240, opened 2016-06-29
Custom config file name/path


fgallese:
Please check the following items before submitting an issue -- thank you!

Note that this project is released with a Contributor Code of Conduct.
By participating in this project you agree to abide by its terms.
[Contributor Code of Conduct](https://voxpupuli.org/coc/).

- [x] There is no existing issue or PR that addresses this problem

Optional, but makes our lives much easier:

- [x] The issue affects the latest release of this module at the time of
  submission

- - -

### What are you seeing

When using a non-default value for either agent or server configuration file, the file it's created, but it is not referenced at the init script. This causes the daemon to use the default values "/etc/zabbix/zabbix_agentd.conf", therefore, the daemon cant start or it starts using the default configuration file which hasn't the custom settings.

### What behaviour did you expect instead

It should set __DAEMON_OPTS="-c [CUSTOMCONFIGFILE]"__ at the init script, so you can use this:

start-stop-daemon --oknodo --start --pidfile $PID --exec $DAEMON **-- $DAEMON_OPTS** >/dev/null 2>&1

At the start function.

### How did this behaviour get triggered

When changing the default values for zabbix config files.

### Output log

N/A

### Any additional information you'd like to impart

N/A

bastelfreak:
I'm just curious, can you tell me why you set a custom config path?

bastelfreak:
#246 got merged

jk2l:
This break 14.04. Systemd is not being used yet in 14.04. The code removed /etc/init.d/zabbix-agent and now the service won't start

fgallese:
@jk2l: I'll add a check for Ubuntu so it takes it into account when detecting the SO, currently it checks "$::osfamily == 'debian'".

jk2l:
Actually i created a patch https://github.com/voxpupuli/puppet-zabbix/pull/252 also done a bit refactoring

fgallese:
@jk2l :+1:

PR #241, opened 2016-06-30
Add config file option to init script

fgallese:
Solution proposed for #240

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -333,7 +333,7 @@ class zabbix::agent (
 		}
 	}
 	
-	if $server_configfile_path != '/etc/zabbix/zabbix_agentd.conf' {
+	if $agent_configfile_path != '/etc/zabbix/zabbix_agentd.conf' {
 		file { '/etc/zabbix/zabbix_agentd.conf':
 			require => Package["$zabbix_package_agent"],
 			ensure  => absent,
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -316,50 +316,50 @@ class zabbix::agent (
   # Ensure that the correct config file is used.
   
   if $::osfamily == 'debian' {
-    if $::operatingsystemrelease !~ /^8.*/ {
+    if $::operatingsystemmajrelease < 8 {
       file { '/etc/init.d/zabbix-agent':
         ensure  => file,
         mode    => '0755',
-        require => Package["$zabbix_package_agent"],
+        require => Package[$zabbix_package_agent],
         content => template('zabbix/zabbix-agent-debian.init.erb'),
       }
     } else {
       file { '/etc/systemd/system/zabbix-agent.service':
         ensure  => file,
         mode    => '0664',
-        require => Package["$zabbix_package_agent"],
+        require => Package[$zabbix_package_agent],
         content => template('zabbix/zabbix-agent-systemd.init.erb'),
       }
       file { '/etc/init.d/zabbix-agent':
         ensure  => absent,
-        require => Package["$zabbix_package_agent"],
+        require => Package[$zabbix_package_agent],
       }
     }
   } elsif $::osfamily == 'redhat' {
-    if $::operatingsystemrelease !~ /^7.*/ {
+    if $::operatingsystemmajrelease < 7 {
       file { '/etc/init.d/zabbix-agent':
         ensure  => file,
         mode    => '0755',
-        require => Package["$zabbix_package_agent"],
+        require => Package[$zabbix_package_agent],
         content => template('zabbix/zabbix-agent-redhat.init.erb'),
       }
     } else {
       file { '/etc/systemd/system/zabbix-agent.service':
         ensure  => file,
         mode    => '0664',
-        require => Package["$zabbix_package_agent"],
+        require => Package[$zabbix_package_agent],
         content => template('zabbix/zabbix-agent-systemd.init.erb'),
       }
       file { '/etc/init.d/zabbix-agent':
         ensure  => absent,
-        require => Package["$zabbix_package_agent"],
+        require => Package[$zabbix_package_agent],
       }
     }
   }
 	
   if $agent_configfile_path != '/etc/zabbix/zabbix_agentd.conf' {
     file { '/etc/zabbix/zabbix_agentd.conf':
-      require => Package["$zabbix_package_agent"],
+      require => Package[$zabbix_package_agent],
       ensure  => absent,
     }
   }
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -312,6 +312,33 @@ class zabbix::agent (
     require => Class['zabbix::repo'],
     tag     => 'zabbix',
   }
+  
+  # Ensure that the correct config file is used.
+  
+  if $::osfamily == 'debian' {
+	file { '/etc/init.d/zabbix-agent':
+		ensure  => present,
+		mode    => '0755',
+		require => Package["$zabbix_package_agent"],
+		replace => true,
+		content => template('zabbix/zabbix-agent-debian.init.erb'),
+	}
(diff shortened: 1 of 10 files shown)


manifests/server.pp
```diff
@@ -423,6 +423,33 @@
...
   }
+  
+  # Ensure that the correct config file is used.
+  
+  if $::osfamily == 'debian' {
+		file { '/etc/init.d/zabbix-server':
+			ensure  => present,
+			mode    => '0755',
+			require => Package["zabbix-server-${db}"],
+			replace => true,
```
juniorsysadmin: This is true by default, so can be removed

manifests/server.pp
```diff
@@ -423,6 +423,33 @@
...
+			content => template('zabbix/zabbix-server-debian.init.erb'),
+		}
+  } elsif $::osfamily == 'redhat' {
+		file { '/etc/init.d/zabbix-server':
+			ensure  => present,
+			mode    => '0755',
+			require => Package["zabbix-server-${db}"],
+			replace => true,
```
juniorsysadmin: This is true by default, so can be removed

manifests/server.pp
fgallese: So, you mean to leave only the _content_ attribute so it always builds the file from the template ?

juniorsysadmin: ```replace => true``` is the default for the File resource

manifests/agent.pp
```diff
@@ -312,6 +312,33 @@
     require => Class['zabbix::repo'],
     tag     => 'zabbix',
   }
+  
+  # Ensure that the correct config file is used.
+  
+  if $::osfamily == 'debian' {
+	file { '/etc/init.d/zabbix-agent':
+		ensure  => present,
```
juniorsysadmin: It's better to make this ```file```rather than ```present```

manifests/agent.pp
```diff
@@ -312,6 +312,33 @@
...
+		mode    => '0755',
+		require => Package["$zabbix_package_agent"],
+		replace => true,
+		content => template('zabbix/zabbix-agent-debian.init.erb'),
+	}
+  } elsif $::osfamily == 'redhat' {
+		file { '/etc/init.d/zabbix-agent':
+			ensure  => present,
```
juniorsysadmin: It's better to make this ```file```rather than ```present```

manifests/server.pp
```diff
@@ -423,6 +423,33 @@
     require => Class['zabbix::repo'],
     tag     => 'zabbix',
   }
+  
+  # Ensure that the correct config file is used.
+  
+  if $::osfamily == 'debian' {
+		file { '/etc/init.d/zabbix-server':
+			ensure  => present,
```
juniorsysadmin: It's better to make this ```file```rather than ```present```

manifests/server.pp
```diff
@@ -423,6 +423,33 @@
...
+			mode    => '0755',
+			require => Package["zabbix-server-${db}"],
+			replace => true,
+			content => template('zabbix/zabbix-server-debian.init.erb'),
+		}
+  } elsif $::osfamily == 'redhat' {
+		file { '/etc/init.d/zabbix-server':
+			ensure  => present,
```
juniorsysadmin: It's better to make this ```file```rather than ```present```

juniorsysadmin:
I'm looking into why the TravisCI tests are failing. It's not because of the code, but your code will fail puppet-lint with those indents.

bastelfreak:
Hi,

could you explain to me why you need this? managing init scripts is always very tricky. I would try to avoid it if possible. Besides the sysv scripts we also need systemd units.

fgallese:
Well, the fact thet you allow to change the path/name of the config path but if you do so it breaks the installation seems off to me.

Personally, I've always used /etc/zabbix/server.conf and /etc/zabbix/agentd.conf, so I changed the value of these in the puppet module. That broke the installation as it didn't start the daemons cause it was looking for the /etc/zabbix/zabbix_agentd.con file.

I saw it as an easy fix, if you change the path/name of the config files I will know it, so why dont I correct the init script so that it doesnt break the installation?

That is the reason why I thought the change was an improvement.

fgallese:
If you are intereses in keeping this change I can add the systemd files.

bastelfreak:
Hi,
please add the systemd files as well. Please take a look at the discussion here: https://github.com/voxpupuli/puppet-kafka/pull/82#discussion-diff-69626042

we should throw modified systemd units into /etc/systemd, not /usr. 

---

Can you please check your configured git email? github is unable to recognize your commits (there isn't your avatar next to your commits).

fgallese:
@bastelfreak I've fixed the github email missconfigurarion. I will now add the systemd as you stated in your message and make a new pull request.

fgallese:
Hello @bastelfreak 

I've added systemd templates on my latest commit.

manifests/agent.pp
```diff
@@ -312,6 +312,57 @@
...
   }
+  
+  # Ensure that the correct config file is used.
+  
+  if $::osfamily == 'debian' {
+    if $::operatingsystemrelease !~ /^8.*/ {
+      file { '/etc/init.d/zabbix-agent':
+        ensure  => file,
+        mode    => '0755',
+        require => Package["$zabbix_package_agent"],
```
bastelfreak: the quotes here aren't needed

manifests/agent.pp
```diff
@@ -312,6 +312,57 @@
     require => Class['zabbix::repo'],
     tag     => 'zabbix',
   }
+  
+  # Ensure that the correct config file is used.
+  
+  if $::osfamily == 'debian' {
+    if $::operatingsystemrelease !~ /^8.*/ {
```
bastelfreak: what happens when debian 9 gets released?

templates/zabbix-agent-systemd.init.erb
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Zabbix Agent
+After=syslog.target network.target
+
+[Service]
+Type=oneshot
```
bastelfreak: this isn't a simple script, zabbix is a daemon. so oneshot is probably wrong. forking should be correct

fgallese: You are right acording to [redhat documentation](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/System_Administrators_Guide/sect-Managing_Services_with_systemd-Unit_Files.html).

I'll fix it along with your other comments.

bastelfreak:
Here are two examples for the unit files:
https://aur.archlinux.org/cgit/aur.git/tree/zabbix-agent.service?h=zabbix3
https://aur.archlinux.org/cgit/aur.git/tree/zabbix-server.service?h=zabbix3

fgallese:
@bastelfreak Thank you. I've reviewed them and after[ checking each parameter ](https://www.freedesktop.org/software/systemd/man/systemd.service.html)I've included them in my latest commit.

bastelfreak:
@fgallese looks good! sorry that this takes so long :(

two things, then we should be done:
the systemd units need to notify systemd. otherwise they arent recognized. I recommend https://github.com/camptocamp/puppet-systemd for this. We already use it in the kafka module.
* add it as a dependency to metadata.json
* add it to .fixtures.yml
* include the class in a sane place
* the file resources that create units should make a notify to the Exec

----

I'm currently doing a modulesync in #245, this should fix your travis issues.

fgallese:
@bastelfreak: Thank you, I've made the changes you mentioned.

bastelfreak:
@fgallese thanks \o/

looks good, could you now please rebase against latest master? Then the tests should turn green and I can merge your PR.

bastelfreak:
￼I am going to close this PR since. The upstream branch is gone and we implemented it already.

PR #246, opened 2016-07-18
Fix for issue #240

fgallese:
Here I'm adding a method to check the zabbix agent/server config file defined by the user.

This fix allaws the user to specify a different config file than the standard /etc/zabbix/zabbix_[agentd|server].conf without breaking the installation.


fgallese:
Hello @bastelfreak: Here I'm creating a new clean pull request with the fixes discussed previously.

bastelfreak:
Hi,
I think you've to use [versioncmp()](https://docs.puppet.com/puppet/latest/reference/function.html#versioncmp) to fix the remaining linter issues.

fgallese:
@bastelfreak: Thank you. I fixed that and it is now passing all tests.

bastelfreak:
\o/

Issue #170, opened 2016-02-17
database is being managed even if manage_database is false


ericsysmin:
Notice: /Stage[main]/Zabbix::Database::Mysql/Exec[zabbix_server_create.sql]/returns: /bin/sh: line 0: cd: /usr/share/doc/zabbix-*-mysql-2.4*/create: No such file or directory
Error: cd /usr/share/doc/zabbix-*-mysql-2.4*/create && if [ -f schema.sql.gz ]; then gunzip schema.sql.gz ; fi && mysql -h 'localhost' -u 'zabbix_server' -p'zabbix' -D 'zabbix' < schema.sql && touch /etc/zabbix/.schema.done returned 1 instead of one of [0]
Error: /Stage[main]/Zabbix::Database::Mysql/Exec[zabbix_server_create.sql]/returns: change from notrun to 0 failed: cd /usr/share/doc/zabbix-*-mysql-2.4*/create && if [ -f schema.sql.gz ]; then gunzip schema.sql.gz ; fi && mysql -h 'localhost' -u 'zabbix_server' -p'zabbix' -D 'zabbix' < schema.sql && touch /etc/zabbix/.schema.done returned 1 instead of one of [0]
Notice: /Stage[main]/Zabbix::Database::Mysql/Exec[zabbix_server_images.sql]: Dependency Exec[zabbix_server_create.sql] has failures: true
Warning: /Stage[main]/Zabbix::Database::Mysql/Exec[zabbix_server_images.sql]: Skipping because of failed dependencies
Notice: /Stage[main]/Zabbix::Database::Mysql/Exec[zabbix_server_data.sql]: Dependency Exec[zabbix_server_create.sql] has failures: true
Warning: /Stage[main]/Zabbix::Database::Mysql/Exec[zabbix_server_data.sql]: Skipping because of failed dependencies
Notice: /Stage[main]/Zabbix::Server/Service[zabbix-server]: Dependency Exec[zabbix_server_create.sql] has failures: true
Warning: /Stage[main]/Zabbix::Server/Service[zabbix-server]: Skipping because of failed dependencies
Notice: Finished catalog run in 0.65 seconds

even though 
zabbix::params::manage_database: false
zabbix::database::manage_database: false

ericsysmin:
i have fixed this in the patch-6 with zabbix::server::manage_database: false now works and stops the errors from occuring

jk2l:
[...]

PR #249, opened 2016-07-22
Fix: manage database parameter not pass to zabbix::server

jk2l:
https://github.com/voxpupuli/puppet-zabbix/issues/170

In issues #170 the parameter was added but it didn't get pass down from zabbix to zabbix::server if using Class['zabbix']


bastelfreak:
tahnks \o/

PR #251, opened 2016-07-22
Add manage_service option. 

Heidistein:
This patch was original from SPBTV/puppet-zabbix:

https://github.com/SPBTV/puppet-zabbix/commit/6013e47e82bc54ecc8ec1b10de668c2c5fe0f2c5
I just made the patch compatible with the main tree again.


Heidistein:
Any chance this will be included? Or feedback what needs to be changed?

bastelfreak:
thanks!

PR #252, opened 2016-07-22
Fix: select systemd for ubuntu correctly

jk2l:
Patch for https://github.com/voxpupuli/puppet-zabbix/issues/240

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -314,51 +314,8 @@ class zabbix::agent (
   }
 
   # Ensure that the correct config file is used.
-  
-  if $::osfamily == 'debian' {
-    if versioncmp($::operatingsystemmajrelease, '8') == -1 {
-      file { '/etc/init.d/zabbix-agent':
-        ensure  => file,
-        mode    => '0755',
-        require => Package[$zabbix_package_agent],
-        content => template('zabbix/zabbix-agent-debian.init.erb'),
-      }
-    } else {
-      include ::systemd
-      file { '/etc/systemd/system/zabbix-agent.service':
-        ensure  => file,
-        mode    => '0664',
-        require => Package[$zabbix_package_agent],
-        content => template('zabbix/zabbix-agent-systemd.init.erb'),
-      } ~>
-      Exec['systemctl-daemon-reload']
-      file { '/etc/init.d/zabbix-agent':
-        ensure  => absent,
-        require => Package[$zabbix_package_agent],
-      }
-    }
-  } elsif $::osfamily == 'redhat' {
-    if versioncmp($::operatingsystemmajrelease, '7') == -1 {
-      file { '/etc/init.d/zabbix-agent':
-        ensure  => file,
-        mode    => '0755',
-        require => Package[$zabbix_package_agent],
-        content => template('zabbix/zabbix-agent-redhat.init.erb'),
-      }
-    } else {
-      include ::systemd
-      file { '/etc/systemd/system/zabbix-agent.service':
-        ensure  => file,
-        mode    => '0664',
-        require => Package[$zabbix_package_agent],
-        content => template('zabbix/zabbix-agent-systemd.init.erb'),
-      } ~>
-      Exec['systemctl-daemon-reload']
-      file { '/etc/init.d/zabbix-agent':
-        ensure  => absent,
-        require => Package[$zabbix_package_agent],
-      }
-    }
+  zabbix::startup {'zabbix-agent':
+    require => Package[$zabbix_package_agent],
   }
 
   if $agent_configfile_path != '/etc/zabbix/zabbix_agentd.conf' {
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -429,51 +429,8 @@ class zabbix::server (
   }
 
   # Ensure that the correct config file is used.
-  
-  if $::osfamily == 'debian' {
-    if versioncmp($::operatingsystemmajrelease, '8') == -1 {
-      file { '/etc/init.d/zabbix-server':
-        ensure  => file,
-        mode    => '0755',
-        require => Package["zabbix-server-${db}"],
-        content => template('zabbix/zabbix-server-debian.init.erb'),
-      }
-    } else {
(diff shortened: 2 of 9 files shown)


bastelfreak:
Maybe we can simplify the logic a bit. We use [camptocamp/systemd](https://github.com/camptocamp/puppet-systemd/) as a dependency, the module provides a [fact](https://github.com/camptocamp/puppet-systemd/blob/master/lib/facter/systemd.rb). We probably should use this instead of parsing operatingsystem versions.

jk2l:
I have done the code changes, and adjusted the tests. But i don't have a system with systemd to test it. so the facter part is solidly base on assumption of the input and output from facter. Need someone help evaluate that section

jk2l:
The refactoring fail:
* First refactoring patch is correct but failed for having quoted boolean word. Maybe we should add puppet-lint skip quoted boolean check?
* Second refactoring failed when run on my server. The facter $::systemd give me 'false' as string which evaluated as true

Feedback require

jk2l:
@bastelfreak who is in charge for the development guideline within this repo?

bastelfreak:
The voxpupuli team is. We define our guidelines in https://github.com/voxpupuli/modulesync_config and they apply to all modules in our namespace. I don't think that we should disable the quoted-boolean checks, bools shouldn't be quoted. I will test the code tomorrow on a systemd box. Facter on puppet3 is a bit painful because it doesn't know any datatypes except for strings :(

bastelfreak:
@jk2l I tested the fact on a CentOS7 and an older Debian box. It correctly returns true/false, but as strings. We have two options here I think:
* Compare the result from `$::systemd` as strings
* cast the string into a bool with [any2bool](https://forge.puppet.com/puppetlabs/stdlib#str2bool)

My personal preference is `any2bool`. Then we should be save even when facter will return the correct datatype.

spec/classes/agent_spec.rb
```diff
@@ -160,4 +161,76 @@
       it { should contain_file('/etc/zabbix/zabbix_agentd.conf').with_content %r{^TLSPSKFile=/etc/zabbix/keys/tlspskfile.key$} }
     end
   end
+
+  context 'On Ubuntu' do
+    context 'trusty' do
+      let :facts do
```
igalic: is there a reason we don't use rspec-facts here?

bastelfreak: this would allow us to shorten the overhead. but then we need am itchy case block to mock the systemd fact depending on the OS :(

jk2l: I only start involve with puppet and open source (only learned rspec last month), so i am still not familiar with the ruby and rspec (PHP background btw).

i can implement the changes, but it will be great if you can show me any reference or guideline that i can follow. as most of the stuff i pick up by reading the github source code directly (it is a bit hard to find the documentation on this)

bastelfreak: rspec-facts is a cool tool which parses the metadata.json and generates a hash of facts for each OS version in the metadata.json.  I already implemented that for a few spec tests: https://github.com/voxpupuli/puppet-zabbix/blob/master/spec/classes/sender_spec.rb#L7-L13

you can always join the #voxpupuli IRC channel on freenode if you've question. direct chat will make it easier for both of us - we're always happy to help.

jk2l:
I never use IRC before. I just joined it but i am not sure did I do it right. it seem like a total empty channel

jk2l:
anyway @bastelfreak i wonder would it be overkill or better if I move the init script section into seperate class and write a unit tests for that instead. which can get rid of duplicate codes in both agent and server section and also the init script coding is not really depends on other system facts.

jk2l:
Finally figured out how to test between different systems and robocop. This should be the last patch

manifests/agent.pp
```diff
@@ -314,16 +314,9 @@
...
-    if versioncmp($::operatingsystemmajrelease, '8') == -1 {
-      file { '/etc/init.d/zabbix-agent':
-        ensure  => file,
-        mode    => '0755',
-        require => Package[$zabbix_package_agent],
-        content => template('zabbix/zabbix-agent-debian.init.erb'),
-      }
-    } else {
+
+  if $::osfamily in ['Debian', 'RedHat'] {
```
bastelfreak: I think we can remove that if? There are multiple other distributions that also ship systemd. the next if with any2bool() should be enough

manifests/agent.pp
```diff
@@ -314,30 +314,8 @@
...
-      $osfamily_downcase = downcase($::osfamily)
-      file { '/etc/init.d/zabbix-agent':
-        ensure  => file,
-        mode    => '0755',
-        require => Package[$zabbix_package_agent],
-        content => template("zabbix/zabbix-agent-${osfamily_downcase}.init.erb"),
-      }
-    }
+  zabbix::startup {'agent':
+    require => Package[$zabbix_package_agent]
```
igalic: i would "hard-code" the title as `zabbix-agent`, so you don't need the variable here
the variable you then only have to pass as `name` to the actual `package` declaration.

manifests/server.pp
```diff
@@ -429,29 +429,8 @@
...
-      file { '/etc/init.d/zabbix-server':
-        ensure  => file,
-        mode    => '0755',
-        require => Package["zabbix-server-${db}"],
-        content => template("zabbix/zabbix-server-${osfamily_downcase}.init.erb"),
-      }
-    }
+  zabbix::startup {'server':
+    require => Package["zabbix-server-${db}"]
   }
```
igalic: same as above wrt agent.

jk2l:
I just realise the any2bool is only available on stdlib master branch. The puppet forge release does not have it yet. Using any2bool also need to move our dependency of stdlib from >= 4.4.0 to newer version that does not exist yet

jk2l:
@bastelfreak Feedback require

bastelfreak:
this may be the reason why the documentation for that is a bit low. I think this is the best function for our usecase. I will see if we can get a release of stdlib.

PR #265, opened 2016-08-18
Restore script /etc/init.d/zabbix-agent from defaults

xdexter:


bastelfreak:
This is one of the reasons why I didn't want to manage the whole file with puppet :(
I'm afraid that we've to check the init script for RHEL/Debian and Zabbix2/3 and find the differences...

xdexter:
Right, i think that better not manage this file too.

bastelfreak:
We need more debugging here. I guess the upstream init scripts for zabbix2 and 3 differ, and probably also the ones for Debian vs RHEL. I'm not sure if I find the time to debug it in the next few days. Maybe someone else can.

bastelfreak:
@xdexter On which RHEL version are you and which zabbix version?

xdexter:
Hello @bastelfreak.

Centos 6.8
Zabbix 2.4.8

HT43-bqxFqB:
On CentOS 6.8 the binary file for the agent is under /usr/sbin/ not under /usr/local/sbin/. Currently there is no way to configure this in the init script. Better option would be to just let the package maintainer handle that and leave the init scripts/unit files as they are installed.

bastelfreak:
Hi @xdexter and @HT43-bqxFqB, we recently introduced this:
https://github.com/voxpupuli/puppet-zabbix/blob/master/templates/zabbix-agent-redhat.init.erb#L21

Does it fix your issues?

xdexter:
Hello @bastelfreak 

Yes, thanks!

Att;

HT43-bqxFqB:
Fixed, thanks!

PR #278, opened 2016-09-20
don't fail if uncompressed file exists

HT43-bqxFqB:
gunzip fails when the uncompressed file exists and requires a user decision which is hidden by the failing puppet run.


bastelfreak:
Hi @HT43-bqxFqB, thanks for the PR. Can you take a look at the failing tests?

HT43-bqxFqB:
The spec test tested for the exact shell command. It's now fixed.

bastelfreak:
Thanks!

Issue #282, opened 2016-09-29
web.pp does not install on Ubuntu 16.04


frozenfoxx:
### Affected Puppet, Ruby, OS and module versions/distributions
- Puppet:  4.7.0
- Ruby:  2.1.9p490
- Distribution:  Ubuntu 16.04
- Module version:  latest on Forge
### How to reproduce (e.g Puppet code you use)

`include zabbix`
### What are you seeing

`E: Package 'php5-pgsql' has no installation candidate`
### What behaviour did you expect instead

Installation of the web component of Zabbix.
### Output log

```
Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install php5-pgsql' returned 100: Reading package lists.
..
Building dependency tree...
Reading state information...
Package php5-pgsql is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'php5-pgsql' has no installation candidate
Error: /Stage[main]/Zabbix::Web/Package[php5-pgsql]/ensure: change from purged to present failed: Execution of '/usr/bin/apt-get -q -y
-o DPkg::Options::=--force-confold install php5-pgsql' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
Package php5-pgsql is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'php5-pgsql' has no installation candidate
```
### Any additional information you'd like to impart

In doing an `apt-cache search` it looks like on Ubuntu 16.04 the `php5-pgsql` package has been replaced with the `php-pgsql` package.  Replacing php5 for php in the event of 16.04 should correct this behavior.

bastelfreak:
thanks for reporting this. are you able to provide a PR to fix it? We should add a case statement to params.pp I guess with a new variable.

frozenfoxx:
I'm still wrestling with understanding how to use modulesync so that's delaying my ability to submit a pull request.  Still, I'll do my best to get one in today that corrects the issue.

frozenfoxx:
As requested a pull request with fixed code has been made [here](https://github.com/voxpupuli/puppet-zabbix/pull/284).

d1nuc0m:
#284 was merged a long ago

PR #284, opened 2016-09-30
Fix for php db package name on Ubuntu 16.04

frozenfoxx:
As explained in [https://github.com/voxpupuli/puppet-zabbix/issues/282](https://github.com/voxpupuli/puppet-zabbix/issues/282) the code for installing the php5 database package does not currently work on Ubuntu 16.04.  This is because the name has changed from "php5-${db}" to "php-${db}" in the official repos.  This pull will update it only in those instances and still use the legacy facts for determining version for backward compatibility.

### DIFF
--- a/spec/classes/web_spec.rb
+++ b/spec/classes/web_spec.rb
@@ -56,7 +56,20 @@ describe 'zabbix::web' do
           let :params do
             super().merge(database_type: 'postgresql')
           end
-          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-pgsql', 'zabbix-web'] : ['zabbix-frontend-php', 'php5-pgsql']
+
+          pgsqlpackage = ""
+          case facts[:osfamily]
+          when 'Debian'
+            if facts[:operatingsystemmajrelease] >= '16.04'
+              pgsqlpackage = "php-pgsql"
+            else
+              pgsqlpackage = "php5-pgsql"
+            end
+          else
+            pgsqlpackage = "php5-pgsql"
+          end
+
+          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-pgsql', 'zabbix-web'] : ['zabbix-frontend-php', pgsqlpackage]
           packages.each do |package|
             it { should contain_package(package) }
           end
@@ -67,7 +80,20 @@ describe 'zabbix::web' do
           let :params do
             super().merge(database_type: 'mysql')
           end
-          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-mysql', 'zabbix-web'] : ['zabbix-frontend-php', 'php5-mysql']
+
+          mysqlpackage = ""
+          case facts[:osfamily]
+          when 'Debian'
+            if facts[:operatingsystemmajrelease] >= '16.04'
+              mysqlpackage = "php-mysql"
+            else
+              mysqlpackage = "php5-mysql"
+            end
+          else
+            mysqlpackage = "php5-mysql"
+          end
+
+          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-mysql', 'zabbix-web'] : ['zabbix-frontend-php', mysqlpackage]
           packages.each do |package|
             it { should contain_package(package) }
           end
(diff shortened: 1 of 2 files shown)


frozenfoxx:
It looks like the `puppet-zabbix/spec/classes/web_spec.rb` lines 59 and 70 are upset with this change since they're looking explicitly for the nonexistent `php5-pgsql` package instead of the new `php-pgsql` package.  I would correct this but I'm not really sure how to rewrite the spec to classify a different package name depending on `facts[:os][:release][:major]` version being >= 16.04.  I'm looking for an example but if someone else can help me rewrite that I think it will pass the tests just fine.

bastelfreak:
Hi @frozenfoxx, thanks for the PR! I guess you have to do something disgusting like I did here:
https://github.com/voxpupuli/puppet-zabbix/blob/master/spec/classes/agent_spec.rb#L17-L22

spec/classes/web_spec.rb
```diff
@@ -56,7 +56,20 @@ def package_provider_for_gems
           let :params do
             super().merge(database_type: 'postgresql')
           end
-          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-pgsql', 'zabbix-web'] : ['zabbix-frontend-php', 'php5-pgsql']
+
+          pgsqlpackage = ""
```
bastelfreak: this is not needed

spec/classes/web_spec.rb
```diff
@@ -67,7 +80,20 @@ def package_provider_for_gems
           let :params do
             super().merge(database_type: 'mysql')
           end
-          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-mysql', 'zabbix-web'] : ['zabbix-frontend-php', 'php5-mysql']
+
+          mysqlpackage = ""
```
bastelfreak: not needed

spec/classes/web_spec.rb
```diff
@@ -56,7 +56,20 @@ def package_provider_for_gems
           let :params do
             super().merge(database_type: 'postgresql')
           end
-          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-pgsql', 'zabbix-web'] : ['zabbix-frontend-php', 'php5-pgsql']
+
+          pgsqlpackage = ""
+          case facts[:osfamily]
+          when 'Debian'
+            if facts[:operatingsystemmajrelease] >= '16.04'
+              pgsqlpackage = "php-pgsql"
```
bastelfreak: migrate all of the double quotes to single quotes unless they contain a variable

spec/classes/web_spec.rb
```diff
@@ -56,7 +56,20 @@ def package_provider_for_gems
           let :params do
             super().merge(database_type: 'postgresql')
           end
-          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-pgsql', 'zabbix-web'] : ['zabbix-frontend-php', 'php5-pgsql']
+
+          pgsqlpackage = ""
+          case facts[:osfamily]
+          when 'Debian'
```
bastelfreak: please migrate to ConditionalAssignment

spec/classes/web_spec.rb
```diff
@@ -67,7 +80,20 @@ def package_provider_for_gems
           let :params do
             super().merge(database_type: 'mysql')
           end
-          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-mysql', 'zabbix-web'] : ['zabbix-frontend-php', 'php5-mysql']
+
+          mysqlpackage = ""
+          case facts[:osfamily]
+          when 'Debian'
+            if facts[:operatingsystemmajrelease] >= '16.04'
```
bastelfreak: same here, please migrate to ConditionalAssignment

spec/classes/web_spec.rb
```diff
@@ -56,7 +56,25 @@ def package_provider_for_gems
           let :params do
             super().merge(database_type: 'postgresql')
           end
-          packages = facts[:osfamily] == 'RedHat' ? ['zabbix-web-pgsql', 'zabbix-web'] : ['zabbix-frontend-php', 'php5-pgsql']
+
+          pgsqlpackage = case facts[:operatingsystem]
```
bastelfreak: almost feels like package maintainer hate us

PR #286, opened 2016-10-04
allow TLS options in Zabbix >= 3.0 and not only = 3.0

cbergmann:


bastelfreak:
Thanks for the PR @cbergmann. Could you add tests for it as well?

bastelfreak:
@cbergmann if you want you can combine this PR with your #287 (if the changes stay in seperate commits).

bastelfreak:
merged https://github.com/voxpupuli/puppet-zabbix/pull/289 which contains the same.

PR #287, opened 2016-10-04
allow TLS options in Zabbix >= 3.0 and not only = 3.0

cbergmann:


bastelfreak:
As mentioned in the other PR, please add tests for it.

bastelfreak:
merged https://github.com/voxpupuli/puppet-zabbix/pull/289 which contains the same

PR #288, opened 2016-10-05
Adjust $pidfile when run on systemd

arioch:

### DIFF
--- a/spec/classes/javagateway_spec.rb
+++ b/spec/classes/javagateway_spec.rb
@@ -5,25 +5,27 @@ describe 'zabbix::javagateway' do
     'rspec.puppet.com'
   end
 
+  facts = {
+    osfamily: 'RedHat',
+    operatingsystem: 'RedHat',
+    operatingsystemrelease: '6.5',
+    operatingsystemmajrelease: '6',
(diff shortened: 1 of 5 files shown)


spec/classes/javagateway_spec.rb
```diff
@@ -5,25 +5,27 @@
     'rspec.puppet.com'
   end
 
+  facts = {
```
bastelfreak: can you migrate this to rspec-puppet-facts?

spec/default_facts.yml
```diff
@@ -4,3 +4,4 @@ ipaddress: "172.16.254.254"
 is_pe: false
 macaddress: "AA:AA:AA:AA:AA:AA"
 selinux_config_mode: "disabled"
+service_provider: 'dummy'
```
bastelfreak: this gets overwritten by our modulesync, please move it to default_module_facts.yml

bastelfreak:
shouldn't /var/run be a symlink to /run? I don't see why the change is needed, could you please explain it?

arioch:
Well this is embarrassing. You're absolutely right. 
I must've misinterpreted the output when checking.

Hence closing.

PR #290, opened 2016-10-09
allow web interface title configuration

HT43-bqxFqB:
Currently the web interface has to have the same IP of the configured server. This makes the window or tab title the IP and not the fqdn. I currently did not add any test to this PR, but will do so if required.
This PR allows one to configure the web with its own name.

### DIFF
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -104,7 +104,11 @@
 #   Database port when not using local socket. Ignored for sqlite.
 #
 # [*zabbix_server*]
-#   The fqdn name of the host running the zabbix-server. When single node:
+#   The ip of the host running the zabbix-server. When single node:
+#   localhost
+#
+# [*zabbix_server_name*]
+#   The fqdn name of the host running the zabbix-server. Shows up in the Webinterface and titleWhen single node:
 #   localhost
 #
 # [*zabbix_listenport*]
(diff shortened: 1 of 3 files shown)


manifests/web.pp
```diff
@@ -104,7 +104,11 @@
 #   Database port when not using local socket. Ignored for sqlite.
 #
 # [*zabbix_server*]
-#   The fqdn name of the host running the zabbix-server. When single node:
+#   The ip of the host running the zabbix-server. When single node:
```
bastelfreak: Please adjust the comment. It defaults to an FQDN, not an IP (I guess the code is correct and the comment is wrong, not the other way round?)

HT43-bqxFqB: Done.

bastelfreak:
Can you please add tests as well? Then I'm happy to merge it.

HT43-bqxFqB:
Will do when I have the time.

bastelfreak:
Hi @HT43-bqxFqB, any progress here? Please also take a look at the used email address in your commit, it isn't associated with your github account.

HT43-bqxFqB:
Ah, thanks. I thought I set it to be not visible. We changed domains recently, so maybe that's why. Regarding the test. That will take some time as I am quite booked. You can decline the PR if you like and I will make a new one when I have some tests can query the web interface in testing.

Issue #264, opened 2016-08-18
zabbix-agentd can't start on CentOS 6.x


xdexter:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.8.7
- Ruby: 1.8.7
- Distribution: CentOS release 6.8 (Final)
- Module version: v2.4.0

### How to reproduce (e.g Puppet code you use)
include zabbix::agent

### What are you seeing
Problem 1: The service zabbix-agent on Centos 6.x have the binary in "/usr/sbin/zabbix_agentd", but template the module (https://github.com/voxpupuli/puppet-zabbix/blob/master/templates/zabbix-agent-redhat.init.erb) use the path "/usr/local/sbin/zabbix_agentd". 

Problem 2: In template init.d (https://github.com/voxpupuli/puppet-zabbix/blob/master/templates/zabbix-agent-redhat.init.erb) have syntax error on "OPTS" variable, the caracter "-c" are in variable (OPTS="-c <%= @agent_configfile_path %>") and daemon (daemon "$ZABBIX_BIN -c $OPTS").

### What behaviour did you expect instead
service start works

### Output log
Problem 1:
Error: Could not start Service[zabbix-agent]: Execution of '/sbin/service zabbix-agent start' returned 5: /usr/local/sbin/zabbix_agentd not installed!
Error: /Stage[main]/Zabbix::Agent/Service[zabbix-agent]/ensure: change from stopped to running failed: Could not start Service[zabbix-agent]: Execution of '/sbin/service zabbix-agent start' returned 5: /usr/local/sbin/zabbix_agentd not installed!

Problem 2:
Iniciando o Zabbix Agent: zabbix_agentd [32380]: cannot open config file [-c]: [2] No such file or directory

### Any additional information you'd like to impart
After correct both problems de service works.

bastelfreak:
ugh :( Thanks for reporting this. Do you think you're able to provide a patch for this?

xdexter:
Hello @bastelfreak,

I think that best solution are use script default from package zabbix-agentd from CentOS, see please: https://github.com/voxpupuli/puppet-zabbix/pull/265. I already tested and working.

xdexter:
Ah, or remove the code in startup.pp:

elsif $::osfamily in ['Debian', 'RedHat'] {
    # Currently other osfamily without systemd is not supported
    $osfamily_downcase = downcase($::osfamily)
    file { "/etc/init.d/${name}":
[...]

PR #291, opened 2016-10-10
Fixes for #264

shaunrampersad:


nibalizer:
Thanks @shaunrampersad !
Looks like this is failing tests, all you need to do is modify https://github.com/voxpupuli/puppet-zabbix/blob/master/spec/defines/startup_spec.rb#L34 a bit.

shaunrampersad:
@nibalizer I've fixed test cases. I've also updated the zabbix-server init as it would also break.

bastelfreak:
Thanks @shaunrampersad!

PR #293, opened 2016-10-11
release 2.6.0

bastelfreak:


bastelfreak:
this release contains an important bugfix, but should not be released before https://github.com/voxpupuli/puppet-zabbix/pull/292 is done.

PR #300, opened 2016-10-18
Add new RPM GPG key for zabbix-3.2 and higher

yevtushenko:
The RPM packages are signed with 082ab56ba14fe591 key:
```
[user@host ~]# rpm -qpi zabbix-agent-3.2.1-1.el6.x86_64.rpm | grep Signature
Signature   : RSA/10, Mon 03 Oct 2016 12:55:23 PM UTC, Key ID 082ab56ba14fe591
[user@host ~]# rpm -qpi zabbix-agent-3.2.1-1.el7.x86_64.rpm | grep Signature
Signature   : RSA/10, Mon 03 Oct 2016 12:55:29 PM UTC, Key ID 082ab56ba14fe591
```

The key installed by puppet-zabbix module isn't correct:
```
[user@host ~]# wget -q http://repo.zabbix.com/RPM-GPG-KEY-ZABBIX -O - | gpg --with-fingerprint
pub  1024D/79EA5ED4 2012-10-28 Zabbix SIA <packager@zabbix.com>
      Key fingerprint = FBAB D5FB 2025 5ECA B22E  E194 D13D 58E4 79EA 5ED4
sub  1024g/7E1DEF85 2012-10-28
```

This one is correct (take a look at last 4 octets of fingerprint):
```
[user@host ~]# wget -q
http://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-A14FE591 -O - | gpg --with-fingerprint
pub  2048R/A14FE591 2016-07-15 Zabbix LLC <packager@zabbix.com>
      Key fingerprint = A184 8F53 52D0 22B9 471D  83D0 082A B56B A14F E591
sub  2048R/E709712C 2016-07-15
```

PR #301, opened 2016-10-19
Added the new ZABIIX-A14FE591 GPG Key.

leemachine:
Zabbix started signing RPMs with a new GPG keypair a few months back. Here is a pull request that keeps the original GPG key but adds the new as well. Without it, Zabbix 3.2 cannot be installed.


bastelfreak:
I'm not sure if the keys have to be listed in an array or as a long string. Have you tested this locally? Or could you maybe add an acceptance test?

bastelfreak:
Hi, thanks for the PR @leemachine, we will go with https://github.com/voxpupuli/puppet-zabbix/pull/300 which implements the same as well.

leemachine:
You're welcome.

Yes, I tested but I like their implementation better. :)

-Brandon

PR #302, opened 2016-10-19
Enhance testing

bastelfreak:


bastelfreak:
This is based on the issues discovered in https://github.com/voxpupuli/puppet-zabbix/pull/300#issuecomment-254795743. Right now the agent_spec is failing on Fedora24 for a single item, I'm not sure why.

PR #306, opened 2016-10-31
Option to define owner of web interface configuration file

admont:
Puppet will fail with errors in case we are not using apache:
Error: Could not find user class ::apache has not been evaluated
Error: Could not find group class ::apache has not been evaluated

for these variables:
$apache_user = getvar('::apache::user')
$apache_group = getvar('::apache::group')

In my commit you can find a possible solution for this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,8 @@ This is the class for installing everything on a single host and thus all parame
 * `zabbix_api_user`: Username of user in Zabbix which is able to create hosts and edit hosts via the zabbix-api. Default: Admin
 * `zabbix_api_pass`: Password for the user in Zabbix for zabbix-api usage. Default: zabbix
 * `zabbix_template_dir`: The directory where all templates are stored before uploading via API
+* `web_config_owner`: Which user should own the web interface configuration file.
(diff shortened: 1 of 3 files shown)


rnelson0:
@admont I cancelled some of the build tasks because they were failing and take a really long time, blocking other builds. If you can, try to run the tests locally first. I think it's the strict_variables that is tripping it up, FWIW.

admont:
@rnelson0, I never used Travis before :-/

bastelfreak:
@admont if you are on a modern ruby version you can do the following locally:
```
bundle install --path .vendor/ --without system_tests development
bundle exec rake test
```

You can get a list of all available tasks by running this:
```
bundle exec rake -T
```

you can run the test task locally and fix the issues it finds. After you committed + pushed them, travis will run the same tests, you don't have to deal directly with travis.

bastelfreak:
If you have any questions to testing or puppet or anything related feel free to join #voxpupuli on freenode.

admont:
I am going to create a new pull request with fixes and own branch in case I need to add other fixes later.

PR #307, opened 2016-11-02
In case we are not using Apache we should have an option to define ow…

admont:
Puppet will fail with errors in case we are not using apache module:
Error: Could not find user class ::apache has not been evaluated
Error: Could not find group class ::apache has not been evaluated

for these variables:
$apache_user = getvar('::apache::user')
$apache_group = getvar('::apache::group')

In my commit you can find a possible solution for this problem.


bastelfreak:
Hi @admont, thanks for the PR. Do you think you can add tests for thisß

bastelfreak:
Hi, could you take a look at the rubocop issues and merge them into your last commit?

admont:
Yes, I will check a bit later, busy with other stuff

bastelfreak:
thanks @admont!

admont:
Thank you @bastelfreak !

Issue #312, opened 2016-11-22
Error: Could not find user class ::apache has not been evaluated


alexjfisher:
With 2.6.0 on EL7, I get the following error on my zabbix_webfrontend role.
```
Error: Could not find user class ::apache has not been evaluated
Error: /Stage[main]/Zabbix::Web/File[/etc/zabbix/web/zabbix.conf.php]/owner: change from apache to class ::apache has not been evaluated failed: Could not find user class ::apache has not been evaluated
Error: Could not find group class ::apache has not been evaluated
Error: /Stage[main]/Zabbix::Web/File[/etc/zabbix/web/zabbix.conf.php]/group: change from apache to class ::apache has not been evaluated failed: Could not find group class ::apache has not been evaluated
```

@bastelfreak Related to https://github.com/voxpupuli/puppet-zabbix/pull/307 perhaps?

alexjfisher:
I think this was probably an issue before https://github.com/voxpupuli/puppet-zabbix/pull/307 too.

alexjfisher:
Buggy version of stdlib's `getvar()` is returning the literal string `'class ::apache has not been evaluated'` instead of `undef`.  Fixed in `stdlib 1.13`.  Code actually tests for empty string, so this might need fixing for puppet 4.

PR #313, opened 2016-11-23
Default web_config_owner/group correctly

alexjfisher:
When apache hasn't been evaluated `getvar('::apache::user')` should
return `undef`.

This commit attempts to address 2 issues.

1) In puppet 4 `''` (empty string) != `undef`
2) In stdlib < 4.13.0 instead of `undef`, `getvar('::apache::user')`
returns a literal string `'class ::apache has not been evaluated'`. This
is a bug that was fixed, but it's not too hard to work around.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -325,31 +325,34 @@ class zabbix::params {
     }
   }
 
+  $default_web_config_owner = $::operatingsystem ? {
+    /(Ubuntu|Debian)/ => 'www-data',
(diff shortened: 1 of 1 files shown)


alexjfisher:
This fixes https://github.com/voxpupuli/puppet-zabbix/issues/312

alexjfisher:
DO NOT MERGE

Testing some simplecov changes

PR #317, opened 2016-11-28
Add file.open() in zabbix_template provider for var template_source v…

DaveHewy:
…alue. As exists? shall never otherwise pass and notify will trigger on each puppet run, which is rather irritating.

### DIFF
--- a/lib/puppet/provider/zabbix_template/ruby.rb
+++ b/lib/puppet/provider/zabbix_template/ruby.rb
@@ -86,7 +86,11 @@ Puppet::Type.type(:zabbix_template).provide(:ruby, parent: Puppet::Provider::Zab
     zabbix_pass = @resource[:zabbix_pass]
     apache_use_ssl = @resource[:apache_use_ssl]
 
+    file = File.open(template_source)
+    template_contents = ''
+    file.each { |line| template_contents << line }
+
     self.class.create_connection(zabbix_url, zabbix_user, zabbix_pass, apache_use_ssl)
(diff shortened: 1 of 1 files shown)


alexjfisher:
@DaveHewy Thanks!

I'm not sure this is the best fix, but it's definitely an improvement and fixes a bug - I wouldn't rule out merging it as-is.

I've had a quick look over the existing code and had the following observations.
* In my opinion `exists?` is doing too much.  Implementing `insync?` in the `ensurable` block in the type would be better.

* There's a [`check_template_exist`](https://github.com/voxpupuli/puppet-zabbix/blob/ed7d06e67dc88d2421c016940e185d6b1263753c/lib/puppet/provider/zabbix.rb#L50) function that looks like it isn't called by anything.  We should either make use of it, or remove it.

DaveHewy:
Hi @alexjfisher agreed.

You do need to be careful that your replacement choice (whether that be my suggestion or not), does still do a comparison on template content, otherwise you will not be able to update templates, only create them if they do not exist. 

In my opinion, the best outcome would be the implementation of both checks, no point opening up that file and attempting to do a comparison if the template does not even exist.

However, my PR just corrects the current implementation. So should go through as a bugfix not a feature.

alexjfisher:
@DaveHewy I'm going to see about adding some unit tests, but so far I've come up with https://github.com/voxpupuli/puppet-zabbix/pull/318

Let me know what you think.

DaveHewy:
@alexjfisher did you get your PR merged in the end for this problem?

alexjfisher:
@DaveHewy Hi! Yes https://github.com/voxpupuli/puppet-zabbix/pull/318 was merged.

PR #318, opened 2016-11-29
Improve zabbix_template type

alexjfisher:
* Implement `insync?` to determine whether a template needs to be
    replaced.
* Refactor connecting to zabbix
* Implement `destroy`
* Remove dead code
* Unit tests for type


DaveHewy:
This makes sense to me. It does take a bug and somewhat re-arrange and add a bunch of new features. Vastly increasing the likelihood of introducing further bugs...

Conceptually though, seems, sound enough.

alexjfisher:
@DaveHewy The more I looked at the old code, the more bugs I saw.  I've added some unit tests to the type, so hopefully we're in a better position going forward.

PR #321, opened 2016-12-01
Replaced agent systemd service with official

### LINKED ISSUE #320: Zabbix agent should not use ProtectSystem in systemd

- Puppet: Any
- Ruby: Any
- Distribution: CentOS 7
- Module version: 2.6.0

class { 'zabbix::agent':
  zabbix_version => '2.2',
  server         => '192.168.1.1',
}

## What are you seeing
systemd service beening replaced with 
```
[Unit]
Description=Zabbix Agent
After=syslog.target network.target
Documentation=man:zabbix_agentd

[Service]
Type=forking
ExecStart=/usr/sbin/zabbix_agentd -c /etc/zabbix/zabbix_agentd.conf
Restart=on-abnormal
PIDFile=/var/run/zabbix/zabbix_agentd.pid
PrivateTmp=yes
ProtectSystem=full
ProtectHome=yes

[Install]
WantedBy=multi-user.target
```
which includes PrivateTmp, ProtectSystem, ProtectHome. 

This options make `vfs.fs.discovery` show mounts that actualy not present, like /home, /etc

Service file should match with package provided file, .
```
[Unit]
Description=Zabbix Agent
After=syslog.target
After=network.target

[Service]
Environment="CONFFILE=/etc/zabbix/zabbix_agentd.conf"
EnvironmentFile=-/etc/sysconfig/zabbix-agent
Type=forking
Restart=on-failure
PIDFile=/run/zabbix/zabbix_agentd.pid
KillMode=control-group
ExecStart=/usr/sbin/zabbix_agentd -c $CONFFILE
ExecStop=/bin/kill -SIGTERM $MAINPID
RestartSec=10s

[Install]
WantedBy=multi-user.target
```

BcTpe4HbIu:
fix #320

### DIFF
--- a/templates/zabbix-agent-systemd.init.erb
+++ b/templates/zabbix-agent-systemd.init.erb
@@ -1,16 +1,16 @@
 [Unit]
 Description=Zabbix Agent
-After=syslog.target network.target
-Documentation=man:zabbix_agentd
+After=syslog.target
+After=network.target
 
 [Service]
 Type=forking
-ExecStart=/usr/sbin/zabbix_agentd -c <%= @agent_configfile_path %>
-Restart=on-abnormal
+Restart=on-failure
 PIDFile=<%= @pidfile %>
-PrivateTmp=yes
-ProtectSystem=full
-ProtectHome=yes
+KillMode=control-group
+ExecStart=/usr/sbin/zabbix_agentd -c <%= @agent_configfile_path %>
+ExecStop=/bin/kill -SIGTERM $MAINPID
+RestartSec=10s
 
 [Install]
 WantedBy=multi-user.target


templates/zabbix-agent-systemd.init.erb
```diff
@@ -1,16 +1,17 @@
 [Unit]
 Description=Zabbix Agent
-After=syslog.target network.target
-Documentation=man:zabbix_agentd
+After=syslog.target
+After=network.target
 
 [Service]
+EnvironmentFile=-/etc/sysconfig/zabbix-agent
```
bastelfreak: this is redhat specific, but we use the unit file also on other systems:
https://github.com/voxpupuli/puppet-zabbix/blob/master/manifests/startup.pp#L52

BcTpe4HbIu: @bastelfreak Removed this line. Anyway there is no /etc/sysconfig/zabbix-agent file even on redhat.

templates/zabbix-agent-systemd.init.erb
```diff
@@ -1,16 +1,16 @@
...
 
 [Service]
 Type=forking
-ExecStart=/usr/sbin/zabbix_agentd -c <%= @agent_configfile_path %>
-Restart=on-abnormal
+Restart=on-failure
 PIDFile=<%= @pidfile %>
-PrivateTmp=yes
-ProtectSystem=full
-ProtectHome=yes
```
bastelfreak: so these additional security settings are pretty smart, but not upstream, I would prefer if we keep them or at least open up a ticket upstream to mention these.

bastelfreak:
thanks!

Issue #323, opened 2016-12-05
Feature Request:  Add more network interface types to listenip


markfaine:
Is it possible to add more network interfaces to the matches for listenip in agent.pp?

I have some VMs that use interfaces that match virbr.* and enp.* but those aren't listed:

```
 if ($listenip != undef) {
    if ($listenip =~ /^(eth|lo|bond|lxc|eno|tap|tun).*/) {
...
```

bastelfreak:
Hey @markfaine, are you able to provide a PR for this?

markfaine:
Yeah sure.

PR #324, opened 2016-12-06
Add virtual bridge (virbr) to the list of network interfaces that are checked when setting listenip

markfaine:
See Issue #323 (Feature Request)

Added virbr interface to list of matched interfaces.  This change allows the IP address to be set from the IP address assigned to a virtual bridge network interface.  

I did not add enp.* since that would require a more substantial modification of the code in order to do correctly since the current matching only matches at the end while predictable network interface names such as enp.* actually match in two locations:  enp[0-9]+s[0-9]+

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -264,7 +264,7 @@ class zabbix::agent (
   # can find the ipaddress of this specific interface if listenip
   # is set to for example "eth1" or "bond0.73".
   if ($listenip != undef) {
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @markfaine, thanks for the PR. Can you please add the used email addr to your profile? github is unable to assign the commit to your account.

markfaine:
@bastelfreak My profile does have an address, though maybe a different address was used for the commit.  I've added that address.  See if it works now.

bastelfreak:
thanks!

Issue #325, opened 2016-12-07
Puppet Unknown variable: '::selinux_config_mode'


alexjfisher:
On EL7, puppet 4.8.1 I get this warning in my puppetserver logs.

bastelfreak:
the PR got merged, thanks!

PR #326, opened 2016-12-07
Strict variables fix for selinux_config_mode

alexjfisher:
Fixes #GH-325

### DIFF
--- a/spec/default_facts.yml
+++ b/spec/default_facts.yml
@@ -3,4 +3,3 @@ concat_basedir: "/tmp"
 ipaddress: "172.16.254.254"
 is_pe: false
 macaddress: "AA:AA:AA:AA:AA:AA"
-selinux_config_mode: "disabled"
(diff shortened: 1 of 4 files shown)


spec/default_facts.yml
```diff
@@ -3,4 +3,3 @@ concat_basedir: "/tmp"
 ipaddress: "172.16.254.254"
 is_pe: false
 macaddress: "AA:AA:AA:AA:AA:AA"
-selinux_config_mode: "disabled"
```
alexjfisher: This was making the tests pass even though on systems with selinux disabled, this fact doesn't exist.

You'd have thought the fact *would* return `disabled`, but if the `selinux` fact is `false`, the `selinux_config_mode` fact isn't returned at all.
The facter 3 code is a bit difficult to read, but in facter 2 there's a [`confine :selinux => true`](https://github.com/puppetlabs/facter/blob/2.4.6/lib/facter/selinux.rb#L137)

alexjfisher: oh really, all modules get this fact set?

PR #354, opened 2017-02-12
release 3.0.0

bastelfreak:


alexjfisher:
The old changelog entries were mostly script generated too? (ie we're not throwing away valuable manually crafted changelog entries).

bastelfreak:
No. I created all of them by hand. I think the now generated ones give us more value, for example links to the PRs/issues.

PR #368, opened 2017-03-16
add config_mode parameter in order to keep some config files private

Lord-Y:


Lord-Y:
Good to go on next release

PR #369, opened 2017-03-17
remove the include ::apt

damoxc:
Having the `include ::apt` in this module then prevents someone from defining their own instance of the apt class with custom parameters in another manifest, forcing the use of hiera to configure the apt class (unless of course this is the intention...).


damoxc:
I guess the better way to handle this would be to include a `manage_apt` option set to true (so as not to interfere with the current behaviour), which then provides someone with a choice.

bastelfreak:
Thanks for the PR @damoxc!

PR #374, opened 2017-04-03
Using $manage_service in zabbix::proxy

Samgarr:
This PR allowing use of $manage_service in zabbix::proxy class.

### DIFF
--- a/manifests/proxy.pp
+++ b/manifests/proxy.pp
@@ -334,6 +334,7 @@ class zabbix::proxy (
   Boolean $manage_firewall         = $zabbix::params::manage_firewall,
   Boolean $manage_repo             = $zabbix::params::manage_repo,
   Boolean $manage_resources        = $zabbix::params::manage_resources,
+  Boolean $manage_service          = $zabbix::params::manage_service,
   $zabbix_proxy                    = $zabbix::params::zabbix_proxy,
   $zabbix_proxy_ip                 = $zabbix::params::zabbix_proxy_ip,
   $use_ip                          = $zabbix::params::proxy_use_ip,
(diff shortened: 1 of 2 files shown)


manifests/proxy.pp
```diff
@@ -542,14 +543,24 @@
...
   # Controlling the 'zabbix-proxy' service
-  service { $proxy_service_name:
-    ensure     => running,
-    hasstatus  => true,
-    hasrestart => true,
-    require    => [
-      Package["zabbix-proxy-${db}"],
-      File[$include_dir],
-      File[$proxy_configfile_path]],
+  if $manage_service == true {
```
bastelfreak: this should be enough:
```
if $manage_service {
```

bastelfreak:
Thanks for the PR @Samgarr. Are you able to also add spec tests for this change?

Samgarr:
I'm not sure right now, Ruby is new for me, spec tests also.

bastelfreak:
Don't worry. Feel free to join our IRC channel #voxpupuli on freenode. We're happy to help you with the tests.

manifests/proxy.pp
```diff
@@ -549,34 +550,54 @@
...
+        database_type     => $database_type,
+        zabbix_type       => 'proxy',
+        database_name     => $database_name,
+        database_user     => $database_user,
+        database_password => $database_password,
+        database_host     => $database_host,
+        zabbix_proxy      => $zabbix_proxy,
+        zabbix_proxy_ip   => $zabbix_proxy_ip,
+        before            => [
+          Service[$proxy_service_name],
```
bastelfreak: is this the only different line? We shouldt do code duplication here. Please write Service[$proxy_service_name] into a variable and use that in the before parameter.

Samgarr:
Rebased, spec tests added.

bastelfreak:
Thanks!

PR #382, opened 2017-05-21
Add acceptance tests + multiple bugfixes

bastelfreak:


bastelfreak:
The diff got quite huge. I initially only wanted to add a single acceptance test, to test the new archlinux vagrant image. I found multiple issues during that. Each important commit has a description which describes the changes.

bastelfreak:
Acceptance tests on vagrant ubuntu16.04 and debian-8 fail, the module isn't idempotent there. It works on fedora-25, fedora-24, archlinux, centos-7,

spec/acceptance/nodesets/archlinux-2-x64.yml
```diff
@@ -0,0 +1,13 @@
+---
```
bastelfreak: I will add this to modulesync_config if this PR gets accepted

spec/spec_helper_acceptance.rb
```diff
@@ -0,0 +1,23 @@
+require 'beaker-rspec'
+require 'beaker/puppet_install_helper'
+require 'beaker/module_install_helper'
+
+run_puppet_install_helper unless ENV['BEAKER_provision'] == 'no'
```
bastelfreak: there are so many different ways to install puppet in a module, is this the current best practice?

Gemfile
```diff
@@ -55,7 +55,7 @@ group :system_tests do
     gem 'beaker-rspec',  :require => false
   end
   gem 'serverspec',                    :require => false
-  gem 'beaker-puppet_install_helper',  :require => false
+  gem 'puppet-module-posix-system-r2.3',  :require => false
```
bastelfreak: Puppetlabs bundled all their gems in a metagem called puppet-module-posix-system. I'm notsure if I really like this or if we should instead include all the gems directly. However, this should be changed in modulesync_config as well.

bastelfreak: also: https://tickets.puppetlabs.com/browse/MODULES-4948 :cry:

spec/spec_helper_acceptance.rb
```diff
@@ -0,0 +1,23 @@
...
+  # Project root
+  proj_root = File.expand_path(File.join(File.dirname(__FILE__), '..'))
+
+  # Readable test descriptions
+  c.formatter = :documentation
+
+  # Configure all nodes in nodeset
+  c.before :suite do
+    hosts.each do |host|
+      copy_module_to(host, :source => proj_root, :module_name => 'ntp')
```
ekohl: I believe the ```install_module_on``` helper already does this for you.

bastelfreak: yep! updated

manifests/agent.pp
```diff
@@ -200,60 +200,64 @@
...
+  $tlscertfile                            = $zabbix::params::agent_tlscertfile,
+  $tlsconnect                             = $zabbix::params::agent_tlsconnect,
+  $tlscrlfile                             = $zabbix::params::agent_tlscrlfile,
+  $tlskeyfile                             = $zabbix::params::agent_tlskeyfile,
+  $tlspskfile                             = $zabbix::params::agent_tlspskfile,
+  $tlspskidentity                         = $zabbix::params::agent_tlspskidentity,
+  $tlsservercertissuer                    = $zabbix::params::agent_tlsservercertissuer,
+  $tlsservercertsubject                   = $zabbix::params::agent_tlsservercertsubject,
+  String $agent_config_owner              = $zabbix::params::agent_config_owner,
+  String $agent_config_group              = $zabbix::params::agent_config_group,
```
oranenj: It's a bit hard to tell here if you added parameters or if this is just whitespace fixing.
It seems agent_config_* are new?

manifests/agent.pp
```diff
@@ -375,4 +380,13 @@
         'ESTABLISHED'],
     }
   }
+  # the agent doesn't work perfectly fine with selinux
+  # https://support.zabbix.com/browse/ZBX-11631
+  if $facts['os']['selinux']['config_mode'] == 'enforcing' {
```
oranenj: Maybe make this conditional on $manage_selinux?

templates/zabbix_agentd.conf.erb
```diff
@@ -5,38 +5,42 @@
...
+<% if @logfile %>
 ### Option: LogFile
 #	Name of log file.
 #	If not set, syslog is used.
 #
-LogFile=<%= @logfile %>
+LogFile=<%= @logfile %><% end %>
 <% end %>
 
 ### Option: LogFileSize
```
apenney: Would it not make sense to wrap the if around the comment block so it doesn't appear if you don't set it?

manifests/agent.pp
bastelfreak: thats why I throw them in their own commit: https://github.com/voxpupuli/puppet-zabbix/pull/382/commits/8bdcaddb246781f8c77d8a008d823e83400dedfd

manifests/agent.pp
bastelfreak: yep, added.

templates/zabbix_agentd.conf.erb
bastelfreak: yep, fixed.

bastelfreak:
If we want to do this the beautiful way we first need to merge https://github.com/voxpupuli/puppet-zabbix/pull/383 :cry:

bastelfreak:
Hi @ianssoftcom, thanks for the notice. I created https://github.com/voxpupuli/puppet-zabbix/pull/396

PR #386, opened 2017-05-22
migrate proxy tests to rspec-puppet-facts

bastelfreak:

### DIFF
--- a/spec/classes/proxy_spec.rb
+++ b/spec/classes/proxy_spec.rb
@@ -5,342 +5,332 @@ describe 'zabbix::proxy' do
     'rspec.puppet.com'
   end
 
-  let :params do
-    {
-      zabbix_server_host: '192.168.1.1',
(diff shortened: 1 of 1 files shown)


spec/classes/proxy_spec.rb
```diff
@@ -5,342 +5,332 @@
...
+          let :params do
+            {
+              manage_resources: true,
+              listenip: '192.168.1.1'
+            }
+          end
+
+          it { is_expected.to contain_class('zabbix::resources::proxy') }
+          it { is_expected.to contain_zabbix__userparameters('Zabbix_Proxy') }
+          it { is_expected.to contain_zabbix__resources__userparameters('foo_Zabbix_Proxy') }
```
alexjfisher: @bastelfreak Where did `foo_Zabbix_Proxy` come from?  Not sure why the tests on this PR all passed, but they don't anymore because of this.

PR #392, opened 2017-05-25
Provide acceptance tests for zabbix server

bastelfreak:


bastelfreak:
As somebody may guess based on the branch name: The initial idea was to add proper acceptance testing in an archlinux docker environment. During that I discovered multiple issues in the existing codebase that also match other operatingsystems.

PR #393, opened 2017-05-31
Execute zabbix-server as zabbix user

krappix:
A good practice is to run a daemon under a non-root user account.


bastelfreak:
Hi @krappix, I'm already working on this in https://github.com/voxpupuli/puppet-zabbix/pull/392

This is a bit more complicated. You probably also don't want to start the agent with the same user as the server. Also there are distributions out there that already provide different users for both services.

bastelfreak:
Hi @krappix, I merged #392, which implements this. I'm going to close this PR, let us know if you need any other changes.

PR #395, opened 2017-06-02
make logtype avaiable to all proxy/server (for 2.6.2)

stanleyz:
logtype was only present as a parameter in agent but not server/proxy

this change combine agent/server/proxy logtype to one parameter in params.pp, and
added parameter to proxy.pp and server.pp

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -53,6 +53,9 @@
 # [*pidfile*]
 #   Name of pid file.
 #
+# [*logtype*]
+#   Where log messages are written to, only available on Zabbix>=3.0
+#
 # [*logfile*]
 #   Name of log file.
 #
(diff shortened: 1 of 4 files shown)


stanleyz:
sorry, this is only for 2.6.2, doesn't seem to be able to create pull request to tags.

bastelfreak:
Hi @stanleyz. This feature is already present in the newer version. Are you able to upgrade your module to a newer version, instead of backporting the features?

stanleyz:
thanks @bastelfreak for response, I am sorry we are still using Puppet 3. 

Besides I don't see logtype parameter in server.pp and proxy.pp of both v4.0.0 and master branch, although it's obviously being used in templates.

bastelfreak:
Hi @stanleyz. I'm going to close this PR due to inactivity. Please reopen it if you're still interested in getting this merged.

PR #396, opened 2017-06-03
only add --foreground on 3.0 and newer

bastelfreak:
We added the --foreground to honor the systemd upstream recommendations.
However, this option is only available in zabbix 3.0 and newer. 2.2
still has support. Additional information:
https://github.com/voxpupuli/puppet-zabbix/pull/382
http://www.zabbix.com/life_cycle_and_release_policy
https://www.zabbix.com/documentation/2.4/manpages/zabbix_agentd
https://www.zabbix.com/documentation/3.0/manpages/zabbix_agentd


bastelfreak:
I just noticed that this patch didn't fix it completely :cry: 

We now provide sane defaults, but it doesn't work if somebody explicitly sets a zabbix version like this:

```puppet
class { 'zabbix::agent':
  server         => '192.168.20.11',
  zabbix_version => '2.4'
}
```

the logic in the params.pp only honors the default value.I will try to spin up a another patch...

Issue #398, opened 2017-06-05
Fails to upgrade zabbix-agent if zabbix_version is changed


bastelfreak:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.10.1
- Ruby: 2.4.1
- Distribution: CentOS
- Module version: master branch

## How to reproduce (e.g Puppet code you use)

Execute this:

```puppet
class{'zabbix-agent':
  zabbix_version => '2.4'
}
```

Than do an upgrade like this:

```puppet
class{'zabbix-agent':
  zabbix_version => '3.2'
}
```

## What are you seeing

zabbix not beeing updated

## What behaviour did you expect instead

zabbix updated to the specified version

## Any additional information you'd like to impart

The package zabbix-agent will never be upgraded since the default behaviour of the package resource is `present`, not `latest`: https://github.com/voxpupuli/puppet-zabbix/blob/master/manifests/params.pp#L71

alexjfisher:
I guess that parameter isn't meant to specify the package version, but is used elsewhere for generating compatible configuration for example?  Maybe add a parameter zabbix_package_version that can default to 'present'?

Issue #397, opened 2017-06-05
Zabbix Repository (zabbix-non-supported) defined with incorrect gpgkey


jamescook000:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.10.1
- Ruby: 2.0.0p648
- Distribution: Centos 7
- Module version: 4.0.0

## How to reproduce (e.g Puppet code you use)

The code used is straight from the modules example:

  class { 'apache':
    mpm_module => 'prefork',
  }
  include apache::mod::php
  class { 'postgresql::server': }
  class { 'zabbix':
    zabbix_version => '3.2',
    zabbix_url        => 'localhost',
  }

## What are you seeing

When trying to install any rpm from the  zabbix-nonsupported repository it fails with the following GPG key error message:

The GPG keys listed for the "Zabbix_nonsupported_7_x86_64" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.

## What behaviour did you expect instead

No errors and Zabbix to install as per module.

## Output log

Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install zabbix-server-pgsql' returned 1: warning: /var/cache/yum/x86_64/7/Zabbix_nonsupported_7_x86_64/packages/fping-3.10-1.el7.x86_64.rpm: Header V4 DSA/SHA1 Signature, key ID 79ea5ed4: NOKEY

The GPG keys listed for the "Zabbix_nonsupported_7_x86_64" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.

 Failing package is: fping-3.10-1.el7.x86_64
 GPG Keys are configured as: http://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-A14FE591
Error: /Stage[main]/Zabbix::Server/Package[zabbix-server-pgsql]/ensure: change from purged to present failed: Execution of '/usr/bin/yum -d 0 -e 0 -y install zabbix-server-pgsql' returned 1: warning: /var/cache/yum/x86_64/7/Zabbix_nonsupported_7_x86_64/packages/fping-3.10-1.el7.x86_64.rpm: Header V4 DSA/SHA1 Signature, key ID 79ea5ed4: NOKEY

The GPG keys listed for the "Zabbix_nonsupported_7_x86_64" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.

 Failing package is: fping-3.10-1.el7.x86_64
[...]

PR #399, opened 2017-06-05
Fix rpm key handling + chaning default values depending on zabbix version

bastelfreak:
These are multiple bugs that I found while doing acceptance testing. The implementation is described in the individual commit messages. This fixes https://github.com/voxpupuli/puppet-zabbix/issues/398 and https://github.com/voxpupuli/puppet-zabbix/issues/397 and is based on the work from https://github.com/voxpupuli/puppet-zabbix/pull/396 and https://github.com/voxpupuli/puppet-zabbix/pull/382

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -68,7 +68,7 @@ class zabbix::params {
   }
 
   # Zabbix overall params. Is used by all components.
-  $zabbix_package_state                     = 'latest'
+  $zabbix_package_state                     = 'present'
   $zabbix_proxy                             = 'localhost'
   $zabbix_proxy_ip                          = '127.0.0.1'
   $zabbix_server                            = 'localhost'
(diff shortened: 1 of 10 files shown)


manifests/params.pp
```diff
@@ -68,7 +68,7 @@
   }
 
   # Zabbix overall params. Is used by all components.
-  $zabbix_package_state                     = 'present'
+  $zabbix_package_state                     = 'latest'
```
alexjfisher: I'm not sure this is a good idea.  Nothing will automatically bump zabbix_version when updated packages get installed.

bastelfreak: this is for the use case that somebody runs zabbix 3.0 or 2.4 and wants to update to 3.2.

robinbowes: Yeah, you should never do this.

alexjfisher: yeah, but this would then just happen whilst zabbix_version would still say 2.4 or whatever.  Incompatible configuration would then be applied.  I think you need a separate parameter for package_version.

PR #413, opened 2017-06-22
Fix proxy service not being enabled

stefanandres:
service enable is not available for RedHat-5.x.
Prior to this commit, the manifest only enabled the service for
\>Redhat-5.x, but not for other operating systems.


bastelfreak:
Thanks @stefanandres!

PR #419, opened 2017-06-26
bump concat to allow 4.X and 3.X

bastelfreak:


bastelfreak:
blargh, we've to wait for this until all dependencies support concat 3.X and newer....

PR #427, opened 2017-07-05
Add "ens*" to interface name matching

nunnun:
Fix an issue with Ubuntu 16.04 TLS having "ens*" names for interfaces

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -293,7 +293,7 @@ class zabbix::agent (
   # can find the ipaddress of this specific interface if listenip
   # is set to for example "eth1" or "bond0.73".
   if ($listenip != undef) {
-    if ($listenip =~ /^(eth|lo|bond|lxc|eno|tap|tun|virbr).*/) {
+    if ($listenip =~ /^(eth|lo|bond|lxc|eno|tap|tun|virbr|ens).*/) {
       $listen_ip = getvar("::ipaddress_${listenip}")
     } elsif is_ip_address($listenip) or $listenip == '*' {
       $listen_ip = $listenip
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -293,7 +293,7 @@ class zabbix::agent (
   # can find the ipaddress of this specific interface if listenip
   # is set to for example "eth1" or "bond0.73".
   if ($listenip != undef) {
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @nunnun, thanks for the PR. Can you take a look at the failing spec tests?

vinzent:
> expected that the catalogue would contain Zabbix::Resources::Userparameters[foo_Zabbix_Proxy]
> https://travis-ci.org/voxpupuli/puppet-zabbix/jobs/250222746#L1460

I don't  see any connection from the chnage to the failed test.

i've restarted the latest master branch job. it  fails too:

https://travis-ci.org/voxpupuli/puppet-zabbix/jobs/247238514#L1424

mmerfort:
Why don't we just do “$listen_ip = getvar("::ipaddress_${listenip}")” every time and then go to the alternatives?

nunnun:
@mmerfort +1 for your idea. I think simply change a code like below

```puppet
  if ($listenip != undef) {
    if is_ip_address($listenip) or $listenip == '*' {
      $listen_ip = $listenip
    } else {
    }
  } else {
    $listen_ip = $::ipaddress
  }
```

nunnun:
@vinzent I added en*,em,wl,ww,p interfaces to the list. But as I mentioned above, I prefer @mmerfort 's idea.

mmerfort:
I don't think falling back to undef is even needed if the default for $listenip is $::ipaddress

```puppet
  if is_ip_address($listenip) or $listenip == '*' {
    $listen_ip = $listenip
  } else {
    unless $listen_ip {
      fail('Good error message')
    }
  }
```

vinzent:
@nunnun As stated I'm not happy with the conditional as it is right now. 

@mmerfort what is returned when `getvar('undefined-var')`is called? Empty string or undef? if empty string then IMHO `unless $listen_ip` isn't doing  what its inteded to.

vinzent:
What is the reasoning at all to allow passing a interface name? If the puppet coder knows the interface he could also pass directly the ip address instead of this interface himself.

bastelfreak:
It may be a good idea to clean this logic up and use it in a profile, if the user really needs it. +1 from my side to the idea from @vinzent.

bastelfreak:
I'm going to close this PR due to inactivity. Please reopen it if you still want to work on it.

PR #433, opened 2017-08-10
Fix default_vhost handling & make web titles configurable 

jvaskonen:
The top-level zabbix class takes a default_vhost parameter, but does not pass it to zabbix::web where it is actually used. Did the same with zabbix_package_state.

Adding the ability to set the title text within the zabbix web UI to a value other than the zabbix_server setting (seeing "localhost" in web titles looks a little odd when connecting to a server by another name.)


manifests/init.pp
```diff
@@ -73,8 +73,10 @@
 class zabbix (
   $zabbix_url                               = '',
   $zabbix_version                           = $zabbix::params::zabbix_version,
+  $zabbix_package_state                     = $zabbix::params::zabbix_package_state,
```
bastelfreak: please add datatypes for those new parameters

bastelfreak:
Hi @jvaskonen. thanks for this PR. Can you please take a look at the used email address in your commits? It isn't associated with your github account. Please also rebase against our latest master branch. This should fix the acceptance tests.

bastelfreak:
Hi. Thanks for raising this PR, but I'm going to close it due to inactivity. Please reopen it if you're still interested in getting this merged.

Issue #334, opened 2016-12-29
Error: Could not set 'present' on ensure: undefined local variable or method `proxy_mode' for #<Puppet::Type::Zabbix_proxy::ProviderRuby:0x00000003aa85b8>


peceka:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.8.1
- Ruby: ruby 2.1.9p490
- Distribution: CentOS7
- Module version: 2.6.1

I've created zabbix proxy with such hiera config:
classes:
  - zabbix::proxy

zabbix::proxy::manage_resources: true
zabbix::mode: 1
zabbix::proxy::mode: 1
zabbix::proxy_mode: 1

so nothing special. Also creating zabbix server was no challenging:
classes:
  - apache::mod::php
  - zabbix
  - postgresql::server

zabbix::manage_resources: true
zabbix::web::manage_resources: true

So when running puppet agent --test on proxy machine it puts data to puppetdb.
And after that running puppet agent --test on zabbix server I receive:

Error: Could not set 'present' on ensure: undefined local variable or method `proxy_mode' for #<Puppet::Type::Zabbix_proxy::ProviderRuby:0x000000042ef6b8>
Error: Could not set 'present' on ensure: undefined local variable or method `proxy_mode' for #<Puppet::Type::Zabbix_proxy::ProviderRuby:0x000000042ef6b8>
Wrapped exception:
undefined local variable or method `proxy_mode' for #<Puppet::Type::Zabbix_proxy::ProviderRuby:0x000000042ef6b8>
Error: /Stage[main]/Zabbix::Resources::Web/Zabbix_proxy[zabbix-proxy.domain.com]/ensure: change from absent to present failed: Could not set 'present' on ensure: undefined local variable or method `proxy_mode' for #<Puppet::Type::Zabbix_proxy::ProviderRuby:0x000000042ef6b8>
Error: Could not set 'present' on ensure: Server answer API error
 {
  "code": -32602,
  "message": "Invalid params.",
  "data": "Trigger has items from template \"Template App Zabbix Proxy\" that is not linked to host."
}
 on request:
 {
  "method": "template.massAdd",
  "params": {
    "hosts": [
      {
        "hostid": null
      }
    ],
    "templates": [
      {
        "templateid": 10048
      }
    ]
  },
  "id": 16118,
  "jsonrpc": "2.0",
  "auth": "57d3992bc5eb3cee2c0f46cdb9efff59"
}
[...]

PR #434, opened 2017-08-10
Fix managing proxy servers

breaker1:
This is a fix for #334 

I found a couple issues with the `zabbix_proxy` ruby code used for managing the proxy resources on the zabbix server. I __think__ this is fixed in a proper way. I am not a Ruby developer, but it has fixed the issues I was experiencing with my manifests.


breaker1:
I don't understand why the unit test is failing... Can anyone help shed some light on this?

bastelfreak:
Hi @breaker1, thanks for this PR. Those issues got fixed in master, could you please rebase against our latest master?

ekohl:
@breaker1 it looks like you merged in master rather than rebase. Usually it's ```git rebase master``` and then you do need to force push but it'll result in a cleaner history and easier for us to review.

breaker1:
@ekohl My apologies. I will try to do it correctly with my next pull request.

ekohl:
@breaker1 no worries, it happens and it's easy to fix.

breaker1:
@ekohl Is that something I need to do? If so, how?

breaker1:
@ekohl @bastelfreak 

Going to close this request and create a new one which is done correctly.

PR #435, opened 2017-08-18
Support multiple zabbix_alias

fribergr:
Enhancement for supporting multiple Alias-directives inside zabbix_agentd.conf.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -235,7 +235,7 @@ class zabbix::agent (
   $buffersend                             = $zabbix::params::agent_buffersend,
   $buffersize                             = $zabbix::params::agent_buffersize,
   $maxlinespersecond                      = $zabbix::params::agent_maxlinespersecond,
-  $zabbix_alias                           = $zabbix::params::agent_zabbix_alias,
+  Optional[Array] $zabbix_alias           = $zabbix::params::agent_zabbix_alias,
   $timeout                                = $zabbix::params::agent_timeout,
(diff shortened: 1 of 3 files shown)


bastelfreak:
Hi @fribergr, thanks for this PR!
* Can you please rebase against our latest master? This should fix the acceptance tests
* Can you please add a few unit tests to verify the file content?

fribergr:
I've added tests and rebased it against master, should be ok now.

juniorsysadmin:
I'm uncomfortable with the abuse of Ruby internals rather than strict typing (requiring the value to be an array), but won't block the change.

bastelfreak:
@juniorsysadmin good catch. Our next release already has a breaking change in it. So should this one here be changed to an array? Or accept string or array as a param in the module, but convert it to an array which gets passed to the template. We do that in other modules as well.

juniorsysadmin:
@bastelfreak I would change it to an array. Less code complexity.

bastelfreak:
@fribergr could you please change the datatype to an array?

fribergr:
@bastelfreak I just removed string support. Does it look ok now?

bastelfreak:
Can you add the datatype Optional[Array] to the zabbix_alias param in agent.pp?

fribergr:
I added the requested Optional[Array], but Travis seems broken (build stuck in queued). Anything else needed?

bastelfreak:
I restarted the job. Secret tip: You can achive the same by closing and reopening the pull request.

PR #442, opened 2017-09-15
modulesync 1.1.0

bastelfreak:


bastelfreak:
fixed the acceptance tests, unit still fails in one case

PR #447, opened 2017-09-21
Add missing proxy_mode variable to zabbix_proxy provider

breaker1:
This is a fix for #334

I found a couple issues with the zabbix_proxy ruby code used for managing the proxy resources on the zabbix server. I think this is fixed in a proper way. I am not a Ruby developer, but it has fixed the issues I was experiencing with my manifests.

### DIFF
--- a/lib/puppet/provider/zabbix_proxy/ruby.rb
+++ b/lib/puppet/provider/zabbix_proxy/ruby.rb
@@ -10,6 +10,10 @@ Puppet::Type.type(:zabbix_proxy).provide(:ruby, parent: Puppet::Provider::Zabbix
     # Set some vars
     host = @resource[:hostname]
     ipaddress = @resource[:ipaddress]
+
+    # Normally 0 is active and 1 is passive, in the API, its 5 and 6
+    proxy_mode = @resource[:mode] + 5
+
     use_ip = @resource[:use_ip]
     port = @resource[:port]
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/zabbix_proxy/ruby.rb
```diff
@@ -10,6 +10,10 @@ def create
     # Set some vars
     host = @resource[:hostname]
     ipaddress = @resource[:ipaddress]
+
+    # Normally 0 is active and 1 is passive, in the API, its 5 and 6
+    proxy_mode = @resource[:mode] + 5
```
ekohl: Why the +5? And does this work because the puppet code passes in a string while the type doesn't munge it to an integer.

breaker1: The comment explains it. I am not sure why, but in the [api documentation ](https://www.zabbix.com/documentation/3.4/manual/api/reference/proxy/create) active proxy is `5` and passive is `6` where in the config it's `0` and `1`. So to make it work I simply add 5 to the value.

bastelfreak:
Hi @breaker1, thanks for the PR. Can you check your mail address here as well?

breaker1:
I have added the email address used to my account.

bastelfreak:
Thanks!

PR #452, opened 2017-09-26
Add support for templated SE Linux agent module

### LINKED ISSUE #441: Allow external script execution with SE Linux
The current SE Linux [module](https://github.com/voxpupuli/puppet-zabbix/blob/master/files/zabbix-agent.te) doesn't seem to be enough to let the Zabbix agent execute scripts.
```bash
allow zabbix_t zabbix_script_exec_t : file { ioctl read getattr execute open } ;
```
And the File resource agent::includedir should probably get the right selinux parameters for the rule to apply.

tux-o-matic:
Fix #441
The current static files used as source for the SE Linux module is too narrow.
With this PR, custom types, classes and rules can be defined for custom SE Linux rules.
For example:
```yaml
zabbix::agent::selinux_require:
  - 'type zabbix_agent_t'
  - 'class process setrlimit'
zabbix::agent::selinux_rules:
  zabbix_agent_t:
    - 'allow zabbix_agent_t self:process setrlimit'
  zabbix_script_t:
    - 'allow zabbix_script_t zabbix_agent_t:process sigchld'
```


manifests/agent.pp
```diff
@@ -415,7 +417,7 @@
   if $facts['selinux'] == true and $manage_selinux {
     selinux::module{'zabbix-agent':
       ensure    => 'present',
-      source_te => 'puppet:///modules/zabbix/zabbix-agent.te',
+      content_te => template('zabbix/selinux/zabbix-agent.te.erb'),
```
bastelfreak: Can you switch this to an epp template instead of erb?

bastelfreak:
Hi @tux-o-matic, thanks for this PR!
* Can you take a look at the failing spec tests?
* Can you add the example for you comment to our README.md?

README.md
```diff
@@ -597,8 +597,21 @@ Many thanks for this!
 ### Standard usage
 *	Not specified as required but for working correctly, the epel repository should be available for the 'fping'|'fping6' packages.
 *	Make sure you have sudo installed and configured with: !requiretty.
-*   Make sure that selinux is permissive or disabled.
 
+### SE Linux
```
bastelfreak: please add a newline after the heading

README.md
```diff
@@ -597,8 +597,21 @@ Many thanks for this!
...
+On systems with SE Linux active and enforcing, Zabbix agent will be limited unless given proper rights with an SE Linux module.
+This Puppet module will apply some default SE Linux rules for it.
+More can be provided if needed by using two class parameters, for example in Hiera YAML:
+```yaml
```
bastelfreak: please add a newline before the codeblock

README.md
bastelfreak: I can recommend https://dlaa.me/markdownlint/ as a linter

tux-o-matic: Actually lint doesn't see that one thing as an issue. And the README is full of it.

tux-o-matic:
There was a failed job in Travis CI due to some internal container or network issue with Travis.

README.md
bastelfreak: yep. github changed their parser in the past, so now we try to enforce their rules for new changes.

bastelfreak:
acceptance tests on centos7 fail. unable to fetch the zabbix-server package. I will take a look later

juniorsysadmin:
Consider adding a parameter for the selinux rules template path, defaulting to selinux/zabbix-agent.te.erb as well.

tux-o-matic:
@juniorsysadmin the agent rules are now a template and the default vales (in params.pp) are the same as the old static files.
And Travis is failing, not the tests.

manifests/agent.pp
```diff
@@ -414,9 +416,9 @@
   # https://support.zabbix.com/browse/ZBX-11631
   if $facts['selinux'] == true and $manage_selinux {
     selinux::module{'zabbix-agent':
-      ensure    => 'present',
-      source_te => 'puppet:///modules/zabbix/zabbix-agent.te',
-      before    => Service['zabbix-agent'],
+      ensure     => 'present',
+      content_te => template('zabbix/selinux/zabbix-agent.te.erb'),
```
juniorsysadmin: @tux-o-matic What I meant is that zabbix/selinux/zabbix-agent.te.erb should also should be the default value of a new parameter named say $selinux_content_te , so if a user doesn't want to use the selinux template included with this module, they can use their own as well.

tux-o-matic: The template is an empty shell, everything is doable via the new class parameters. This PR only adds flexibility.

bastelfreak:
@tux-o-matic can you please rebase against our latest master?

tux-o-matic:
@bastelfreak rebased and green CI.

bastelfreak:
It looks like you didn't rebase, but merged master into this branch, but that is okay as well.

PR #454, opened 2017-09-28
wait for network connectivity before Zabbix Server startup

tequeter:
Zabbix Server starts too early in the boot process :
- It can't reach its database (if remote).
- It may have missing data for all the items that require network access.

This change makes it wait until routed networking is available.

### DIFF
--- a/templates/zabbix-server-systemd.init.erb
+++ b/templates/zabbix-server-systemd.init.erb
@@ -6,6 +6,10 @@ After=mysqld.service
 <% elsif @database_type == "postgresql" and @manage_database %>
 After=postgresql.service
 <% end -%>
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
Unsure why the test is failing.

Just a comment: Ideally all template paths used to populate init scripts should be a parameter so users can use their own (In this case the default should be `$server_systemd_init_script_file_template = 'zabbix/zabbix-server-systemd.init.erb'` for example).

bastelfreak:
Hi @tequeter, thanks for this PR. Are you able to rebase against our latest master branch? This fixes the acceptance tests

tequeter:
There, rebased. Better late than never :)

PR #456, opened 2017-10-11
release 5.0.1

bastelfreak:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,25 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+## [v5.0.1](https://github.com/voxpupuli/puppet-zabbix/tree/v5.0.1) (2017-10-21)
+[Full Changelog](https://github.com/voxpupuli/puppet-zabbix/compare/v5.0.0...v5.0.1)
+
+**Fixed bugs:**
+
+- add selinux workaround for IPC in zabbix-server3.4 [\#459](https://github.com/voxpupuli/puppet-zabbix/pull/459) ([bastelfreak](https://github.com/bastelfreak))
+- Remove updateExisting from applications in zabbix template provider. [\#450](https://github.com/voxpupuli/puppet-zabbix/pull/450) ([breaker1](https://github.com/breaker1))
+- Add missing proxy\_mode variable to zabbix\_proxy provider [\#447](https://github.com/voxpupuli/puppet-zabbix/pull/447) ([breaker1](https://github.com/breaker1))
+
+**Closed issues:**
+
+- Zabbix API errors when adding Templates in Zabbix 3.4 v2 [\#449](https://github.com/voxpupuli/puppet-zabbix/issues/449)
+- Zabbix API errors when adding Templates in Zabbix 3.4 [\#446](https://github.com/voxpupuli/puppet-zabbix/issues/446)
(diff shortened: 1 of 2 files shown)


bastelfreak:
https://repo.zabbix.com/zabbix/3.2/rhel/7/x86_64/zabbix-server-pgsql-3.2.8-1.el7.x86_64.rpm is available on the mirrors, but there is also a newer patch release. I am not sure about the root cause here.

wyardley:
@ekohl @bastelfreak I think I looked into this earlier and suggested a fix on IRC. I believe the actual root cause is that things don't get reset between the various tests. I think you just need to adjust the test to remove the package between versions, similar to what we do when testing installing packages from different sources in the nginx module.

wyardley:
So, that is a partial fix....
```
--- a/spec/acceptance/server_spec.rb
+++ b/spec/acceptance/server_spec.rb
@@ -20,6 +20,10 @@ describe 'zabbix::server class' do
         class { 'zabbix::server': }
       EOS
 
+      if fact('os.family') == 'RedHat'
+        shell('yum clean metadata')
+      end
+
       # Run it twice and test for idempotency
       apply_manifest(pp, catch_failures: true)
       apply_manifest(pp, catch_changes: true)
```
fixes the error with the postgresql package.

However, the second run is still exiting 2. It looks like the first run tries to ensure running:
```
  Notice: /Stage[main]/Zabbix::Server/Service[zabbix-server]/ensure: ensure changed 'stopped' to 'running'
  Info: /Stage[main]/Zabbix::Server/Service[zabbix-server]: Unscheduling refresh on Service[zabbix-server]
```
but the second run also has this.
```
  Notice: /Stage[main]/Zabbix::Server/Service[zabbix-server]/ensure: ensure changed 'stopped' to 'running'
  Info: /Stage[main]/Zabbix::Server/Service[zabbix-server]: Unscheduling refresh on Service[zabbix-server]
  Notice: Applied catalog in 2.76 seconds
```

If I replicate this condition, I see this status in between the two runs:
```
Redirecting to /bin/systemctl status zabbix-server.service
● zabbix-server.service - Zabbix Server
   Loaded: loaded (/etc/systemd/system/zabbix-server.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Fri 2017-10-20 05:55:15 UTC; 14s ago
     Docs: man:zabbix_server
 Main PID: 8439 (code=exited, status=0/SUCCESS)

Oct 20 05:55:13 centos-7-x64 systemd[1]: Started Zabbix Server.
Oct 20 05:55:13 centos-7-x64 systemd[1]: Starting Zabbix Server...
Oct 20 05:55:13 centos-7-x64 zabbix_server[8439]: Starting Zabbix Server. Zabbix 3.4.3 (revision 73588).
Oct 20 05:55:13 centos-7-x64 zabbix_server[8439]: Press Ctrl+C to exit.
```
It looks like a successful start, and the error sometimes keeps occurring. Will run the test one more time to see if I can figure out what the error is.

wyardley:
So even though the test claims zabbix-server is running (probably because it hasn't died yet), the actual error is this:

```
  7146:20171020:061015.337 server #0 started [main process]
  7180:20171020:061015.357 server #28 started [icmp pinger #1]
  7182:20171020:061015.357 server #30 started [preprocessing manager #1]
  7182:20171020:061015.357 cannot start preprocessing service: Cannot bind socket to "/tmp/zabbix_server_preprocessing.sock": [13] Permission denied.
  7146:20171020:061015.359 One child process died (PID:7182,exitcode/signal:1). Exiting ...
```
file permissions in `/tmp` are ok, so I think selinux is the issue
https://support.zabbix.com/browse/ZBX-12567

PR #458, opened 2017-10-20
Fix dependency cycle with Package[dirmngr]

remisauvat:
When ::apt is included by ::zabbix::repo it gets automaticaly tagged with zabbix.
So we need to use a custom tag to set the correct order for packages manage by the zabbix module.
Else we get a dependency cycle with the package dirmngr installed by ::apt in recent Debian 9 or Ubuntu

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -339,7 +339,7 @@ class zabbix::agent (
   package { $zabbix_package_agent:
     ensure  => $zabbix_package_state,
     require => Class['zabbix::repo'],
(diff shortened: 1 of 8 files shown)


bastelfreak:
HI @rsauvatinet, thanks for this PR. Are you able to rebase against our latest master? This will fix some issues in the tests.

bastelfreak:
Hi @remisauvat. I'm going to close this PR due to inactivity. Please feel free to reopen it if you still want it merged.

PR #465, opened 2017-11-16
modulesync 1.5.0

bastelfreak:


.sync.yml
```diff
@@ -1,30 +1,10 @@
...
-    sudo: required
-    bundler_args: --without development
-    dist: trusty
-  - rvm: 2.4.1
-    env: PUPPET_INSTALL_TYPE=agent CHECK=acceptance BEAKER_debug=true BEAKER_set=docker/ubuntu-16.04
-    services: docker
-    sudo: required
-    bundler_args: --without development
-    dist: trusty
+  docker_sets:
```
bastelfreak: I am not sure if this syntax is correct

TraGicCode: A quick modulesync locally should be able to figure this out one second.

TraGicCode: Fixed it @bastelfreak

PR #467, opened 2017-11-17
Remove EOL operatingsystems

ekohl:


metadata.json
```diff
@@ -144,9 +144,7 @@
     },
     {
       "operatingsystem": "Fedora",
-      "operatingsystemrelease": [
-        "24"
-      ]
+      "operatingsystemrelease": []
```
bastelfreak: this means that we support all versions of Fedora, right?

ekohl: Not sure. I honestly would like to keep Fedora support in but without acceptance tests I wouldn't want to claim that.

bastelfreak: Yeah I would like to keep it as well. We got a bunch of module where be claim support for something just because rspec-puppet-facts doesn't throw an error. We don't have acceptance tests for all operating systems. Did you try adding fedora25/26 to the metadata.json and check if it still compiles?

juniorsysadmin: I'm okay with leaving a broad Fedora entry in there as long as there is some kind of entry for it under https://github.com/voxpupuli/puppet-nodejs#limitations describing that it _should_ work, is best effort supported etc

Issue #473, opened 2018-01-04
zabbix::agent without an IP shouldn't guess the IP


tibers:
https://github.com/voxpupuli/puppet-zabbix/blob/master/manifests/agent.pp#L293-L307

This is bad because if we're using this module when building containers for docker/kubernetes/etc, the module will grab the temporary IP of the VM in the build service, which will be incorrect when the container is deployed one or more times into a production environment. 
  
Omitting the ListenIP line entirely makes the agent pick whatever IP it thinks is appropriate, so it's a much cleaner solution.

bastelfreak:
Hi @tibers, thanks for rising this issue. Are you able to create a pull request with the needed changes?

tibers:
I'm not entirely sure I understand the code enough to make this happen but I'll give it a go. :)

tibers:
Opened [this PR](https://github.com/voxpupuli/puppet-zabbix/pull/474), hopefully I didn't goose things up too badly....

Helenofpern:
Any news on when this will be implemented as it is a right pain when managing aws instances across many accounts.

bastelfreak:
This should be fixed with https://github.com/voxpupuli/puppet-zabbix/pull/487

PR #474, opened 2018-01-05
making listenip optional for the zabbix agent

tibers:
Addressing [Issue 473](https://github.com/voxpupuli/puppet-zabbix/issues/473) concerning the ListenIP being optional in the zabbix-agent config. Makes things much smoother for users installing zabbix into an image generated with packer, etc.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -85,8 +85,7 @@
 #   listenip => 'eth0',  or
 #   listenip => 'bond0.73',
 #
-#   It will find out which ip is configured for this ipaddress. Can be handy
-#   if more than 1 interface is on the server.
+#   If left blank, ListenIP will not be in the configuration file and and the agent will pick.
 #
 # [*startagents*]
 #   Number of pre-forked instances of zabbix_agentd that process passive checks.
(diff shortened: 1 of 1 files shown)


manifests/agent.pp
```diff
@@ -303,7 +302,7 @@
       $listen_ip = $::ipaddress
     }
   } else {
-    $listen_ip = $::ipaddress
+    notify { 'listenip is undef and will not be written to the configuration file' : }
```
bastelfreak: this is not a good idea, because this will show up in each run as an applied resource. We should probably leave the variable undefined or empty string, and check for that in the template.

bastelfreak:
Can you add a test that verifies that we don't write the empty argument into the config file?

tibers:
OK that's actually where I get sketchy. I can do puppet code, but I don't know how to write tests. :\

bastelfreak:
Hi @tibers, one example would be this: https://github.com/voxpupuli/puppet-zabbix/blob/master/spec/classes/agent_spec.rb#L204
Feel free to join our IRC channel #voxpupuli on freenode. We're always happy to assist you in writing tests!

tibers:
We abandoned zabbix for sensu. Thanks for the tip though.

Issue #476, opened 2018-01-11
Support HTTP auth for API calls


tux-o-matic:
The [Zabbix API Ruby client](https://github.com/express42/zabbixapi/blob/c5bfd135796797adc335989322d4e784c61eeb6b/lib/zabbixapi/client.rb#L101) supports Zabbix authentication and HTTP auth (handled by Apache).
However this module only uses the Zabbix user and password when [calling the API](https://github.com/voxpupuli/puppet-zabbix/blob/5207b00cdaef1833a005874ffe85e062b9460d42/lib/puppet/provider/zabbix.rb#L11).

In situations where HTTP auth is enabled in Apache and Zabbix, the Puppet module can no longer interact with the API. To present such situations, this module should pass a user and password value for HTTP authentication if set to do so.

PR #479, opened 2018-01-19
#476 Pass credentials for HTTP auth

tux-o-matic:
Along the usual Zabbix user and password, the credentials are now passed on to the Zabbix api client for HTTP basic auth support.

### DIFF
--- a/lib/puppet/provider/zabbix.rb
+++ b/lib/puppet/provider/zabbix.rb
@@ -11,7 +11,9 @@ class Puppet::Provider::Zabbix < Puppet::Provider
     zbx = ZabbixApi.connect(
       url: "#{protocol}://#{zabbix_url}/api_jsonrpc.php",
       user: zabbix_user,
-      password: zabbix_pass
+      password: zabbix_pass,
+      http_user: zabbix_user,
+      http_password: zabbix_pass
     )
     zbx
   end


tux-o-matic:
There is one test failing and it seems to be related to Travis and/or beaker.

bastelfreak:
restarted the failed job

bastelfreak:
Fixes https://github.com/voxpupuli/puppet-zabbix/issues/476

lib/puppet/provider/zabbix.rb
```diff
@@ -11,7 +11,9 @@ def self.create_connection(zabbix_url, zabbix_user, zabbix_pass, apache_use_ssl)
     zbx = ZabbixApi.connect(
       url: "#{protocol}://#{zabbix_url}/api_jsonrpc.php",
       user: zabbix_user,
-      password: zabbix_pass
+      password: zabbix_pass,
+      http_user: zabbix_user,
```
bastelfreak: what happens when those variables are empty?

tux-o-matic: They never are empty since they re use the same regular Zabbix API credentials.
And in a setup without HTTP auth, the HTTP Auth header is just ignored by Apache.

bastelfreak: okay, thanks for the explanation.

Issue #477, opened 2018-01-15
SELinux blocks httpd from accessing database


olifre:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.3.3
- Ruby: (builtin since puppet 5)
- Distribution: CentOS 7.4
- Module version: 480541e315be75fe6ecba83031fa04f1157c890e

## How to reproduce (e.g Puppet code you use)
Most basic setup, PostgreSQL-DB, Zabbix-Server and Web-UI one a single host. 

## What are you seeing
SELinux complains:
```
type=AVC msg=audit(1516033832.022:66562): avc:  denied  { name_connect } for  pid=10101 comm="httpd" dest=5432 scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:object_r:postgresql_port_t:s0 tclass=tcp_socket
```
and Web-UI does not work. 

## What behaviour did you expect instead
I expected the puppet module to set:
```
setsebool httpd_can_network_connect_db=1
```
which fixes this.

Fabian1976:
I ran into the same issue. And created a PR #481

tux-o-matic:
It shouldn't be up to this module to set SE Linux for this. Should be the Zabbix RPM.

Fabian1976:
Yes, it should. But it's not.
In the PR, you can see that we decided NOT to included this SEBoolean in the module but we did make a note of it in the README.
It should be done by the RPM, but now we advise to apply it in a profile.

bastelfreak:
As noticed by @Fabian1976, this is expected behaviour. I'm going to close this issue. @olifre please reopen if you think we should further discuss this.

PR #481, opened 2018-01-23
Document needed sebooleans for httpd/zabbix-web

Fabian1976:
The web.pp missed these sebooleans:

- httpd_can_network_connect
- httpd_can_network_connect_db

As mentioned in issue #477

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,9 +170,9 @@ Everything will be installed on the same server. There is also an possibility to
 https://github.com/voxpupuli/puppet-zabbix/wiki/Multi-node-Zabbix-Server-setup
 
 Please note that if you use apache as the frontend (which is the default) and SELinux is enabled, you need to set these SEBooleans (preferably in a profile) to allow apache to connect to the database:
-```ruby
+```puppet
 if $facts['selinux'] {
-  selboolean { ['httpd_can_network_connect', 'httpd_can_network_connect_db']
+  selboolean { ['httpd_can_network_connect', 'httpd_can_network_connect_db']:
     persistent => true,
     value      => 'on',
   }
--- a/README.md
+++ b/README.md
@@ -169,6 +169,16 @@ node 'zabbix.example.com' {
 Everything will be installed on the same server. There is also an possibility to seperate the components, please check the following wiki:
 https://github.com/voxpupuli/puppet-zabbix/wiki/Multi-node-Zabbix-Server-setup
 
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @Fabian1976, thanks for this PR. I'm not sure if we should include those sebooleans in the module. They are apache specific, not zabbix specific. It is possible, that others manage this seboolean already, but outside of the module. Maybe it is better to handle this in a profile and add a note to the readme?

ekohl:
I wonder if ```ensure_resource``` should be used to handle possible duplicate definitions. This is always hard and IMHO an unsolved problem.

Fabian1976:
You could also fix this in a profile. But if you include zabbix::web (which needs apache), people assume that it sets the appropiate SEbooleans (it sets 1, why not all that are needed for web to work?)

bastelfreak:
There is a small but important difference. It sets all needed sebooleans that httpd needs to talk to zabbix. It doesn't  set all sebooleans that httpd need. IMO we shouldn't manage those sebooleans in the zabbix module. They should be part of httpd module or a profile.

Fabian1976:
I can live with that. I allready added it to a profile as a work around. But then i will make it permanent. Not a problem. Shall i close my PR?

ekohl:
It would be great if it's at least documented in the README.

bastelfreak:
@Fabian1976 can you remove the current changes and instead document it in the README.md?

Fabian1976:
Yes, not a problem. Will do

Fabian1976:
I updated the README

README.md
```diff
@@ -169,6 +169,16 @@ node 'zabbix.example.com' {
 Everything will be installed on the same server. There is also an possibility to seperate the components, please check the following wiki:
 https://github.com/voxpupuli/puppet-zabbix/wiki/Multi-node-Zabbix-Server-setup
 
+Please note that if you use apache as the frontend (which is the default) and SELinux is enabled, you need to set these SEBooleans (preferably in a profile) to allow apache to connect to the database:
+```ruby
```
ekohl: There's puppet highlighting too

README.md
```diff
@@ -169,6 +169,16 @@ node 'zabbix.example.com' {
...
+if $facts['selinux'] {
+  selboolean { ['httpd_can_network_connect', 'httpd_can_network_connect_db']
```
ekohl: Missing a colon here

Fabian1976: Fixed it

PR #482, opened 2018-01-24
add vmwaretimeout as possible server-param

b014651:
Due to error-messages in my zabbix_server.log file, I have to increase the vmwaretimeout option.
This parm was already defined in the zabbix_server.conf.erb - Template, but was not available/referenced as parm for zabbix_server configuration.  
Added the possibility to add vmwaretimeout parm


bastelfreak:
Thanks for this PR @b014651 !

PR #486, opened 2018-03-13
allow camptocamp/systemd 2.X

bastelfreak:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -42,7 +42,7 @@
     },
     {
       "name": "camptocamp/systemd",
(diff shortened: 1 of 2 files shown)


bastelfreak:
funfact: rspec-puppet-fact had no facts for archlinux, so we didn't test it. It got recently added and revealed a bug in the mysql module

bastelfreak:
proposed fix: https://github.com/puppetlabs/puppetlabs-mysql/pull/1057

bastelfreak:
rebased

PR #487, opened 2018-03-13
Do not guess ListenIP, fixes #473, bump stdlib to 4.19.0

### LINKED ISSUE #473: zabbix::agent without an IP shouldn't guess the IP

lordievader:
This commit changes the behaviour if the $listen_ip argument is not
given. In that case ListenIP is omitted from the agent configuration.
Having no ListenIP in the configuration ensures the agent listens on
both IPv4 and IPv6.
Also added a spec test for the case were ListenIP is left empty.
Acceptance testing gave no errors.

Signed-off-by: Olivier van der Toorn <oliviervdtoorn@gmail.com>

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -294,16 +294,10 @@ class zabbix::agent (
   # to network name. If more than 1 interfaces are available, we
   # can find the ipaddress of this specific interface if listenip
   # is set to for example "eth1" or "bond0.73".
-  if ($listenip != undef) {
-    if ($listenip =~ /^(eth|lo|bond|lxc|eno|tap|tun|virbr).*/) {
-      $listen_ip = getvar("::ipaddress_${listenip}")
-    } elsif is_ip_address($listenip) or $listenip == '*' {
-      $listen_ip = $listenip
-    } else {
-      $listen_ip = $::ipaddress
-    }
-  } else {
-    $listen_ip = undef
+  $listen_ip = $listenip ? {
+    /^(eth|lo|bond|lxc|eno|tap|tun|virbr).*/ => fact("networking.interfaces.${listen_ip}.ip"), # the stdlib fact function still returns undef if the fact can't be found.  This keeps the behaviour similar to how getvar would have worked, but now we use the modern networking fact instead of the legacy ipaddress_INTERFACE one
+    '*' => undef,
+    default => $listenip, # I think this is then right for all other possibilities (an ip address, or undef)
   }
 
   # So if manage_resources is set to true, we can send some data
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -303,7 +303,7 @@ class zabbix::agent (
       $listen_ip = $::ipaddress
     }
   } else {
-    $listen_ip = $::ipaddress
+    $listen_ip = undef
   }
 
   # So if manage_resources is set to true, we can send some data
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -295,9 +295,9 @@ class zabbix::agent (
   # can find the ipaddress of this specific interface if listenip
   # is set to for example "eth1" or "bond0.73".
   $listen_ip = $listenip ? {
-    /^(eth|lo|bond|lxc|eno|tap|tun|virbr).*/ => fact("networking.interfaces.${listen_ip}.ip"), # the stdlib fact function still returns undef if the fact can't be found.  This keeps the behaviour similar to how getvar would have worked, but now we use the modern networking fact instead of the legacy ipaddress_INTERFACE one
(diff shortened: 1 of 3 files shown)


alexjfisher:
@lordievader Thanks!  Do you think the whole section can be greatly simplified?  The nested if's are quite hard to read.

From
```puppet
if ($listenip != undef) {
    if ($listenip =~ /^(eth|lo|bond|lxc|eno|tap|tun|virbr).*/) {
    } elsif is_ip_address($listenip) or $listenip == '*' {
      $listen_ip = $listenip
    } else {
      $listen_ip = $::ipaddress
    }
  } else {
    $listen_ip = $::ipaddress
  }
```
to something similar to...
```puppet
# change to parameter definition.
Optional[Variant[Pattern[/^(eth|lo|bond|lxc|eno|tap|tun|virbr).*/,/^\*$/],Stdlib::Ip_address]] $listenip = $zabbix::params::agent_listenip,

# then the code becomes
$listen_ip = $listenip ? {
  /^(eth|lo|bond|lxc|eno|tap|tun|virbr).*/ => fact("networking.interfaces.${listen_ip}.ip"), # the stdlib fact function still returns undef if the fact can't be found.  This keeps the behaviour similar to how getvar would have worked, but now we use the modern networking fact instead of the legacy ipaddress_INTERFACE one
  default => $listenip, # I think this is then right for all other possibilities ('*', an ip address, or undef)
}
```

alexjfisher:
https://github.com/voxpupuli/puppet-zabbix/blob/master/metadata.json#L13 would need updating slightly.  [fact()](https://github.com/puppetlabs/puppetlabs-stdlib#fact) was introduced in [4.19.0](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/CHANGELOG.md#supported-release-4190)

bastelfreak:
@alexjfisher thanks for the work. The new datatype for $zabbix::params::agent_listenip should be put into its own type.

alexjfisher:
@bastelfreak Yeah, I was thinking about that.  It is a big long.

lordievader:
@alexjfisher Looks alot nicer, I agree. However, the `ListenIP` clause should be omitted if the parameter is not given. With a '*' or '::' you get errors like:
`cannot send list of active checks to "<some-ip>": "*" is not a valid IP address`

alexjfisher:
@lordievader So the original code is broken for '*' ?

```puppet
elsif is_ip_address($listenip) or $listenip == '*' {
      $listen_ip = $listenip
    }
```

lordievader:
No, this is more strange behaviour of the Zabbix agent config. With '*' the Zabbix server doesn't want to send active checks. Without the `ListenIP` clause in the config the agent listens to any IPv4 (`*:10050`) and any IPv6 address (`:::10050`). Which is, I think, the intended behaviour of '*' in the config.

alexjfisher:
So it should be
```puppet
$listen_ip = $listenip ? {
  /^(eth|lo|bond|lxc|eno|tap|tun|virbr).*/ => fact("networking.interfaces.${listen_ip}.ip"), # the stdlib fact function still returns undef if the fact can't be found.  This keeps the behaviour similar to how getvar would have worked, but now we use the modern networking fact instead of the legacy ipaddress_INTERFACE one
  '*' => undef,
  default => $listenip, # I think this is then right for all other possibilities (an ip address, or undef)
}
```
?

lordievader:
Yes, if I read it correctly as a switch statement. This has the intended behaviour. :+1:
I can change the commit in the afternoon and change the tests to reflect this change.

alexjfisher:
Similar to a switch, it's a selector. https://puppet.com/docs/puppet/5.3/lang_conditional.html#selectors

lordievader:
Selector replacement commited!

alexjfisher:
Great! Does it work? (I never tested it).
Also, the end of line comments were for you. I don’t think we want to actually have them in the finished code. :)

lordievader:
Yes, it works. Modified the test to double-check.

bastelfreak:
@lordievader can you please rebase and take a look at the rubocop issues?

bastelfreak:
I rebased the branch

alexjfisher:
Oops.  My fault, but there was a slight bug in this PR.
See https://github.com/voxpupuli/puppet-zabbix/pull/494

PR #489, opened 2018-03-27
bump mysql version to >= 5.4.0 < 6.0.0

bastelfreak:
This is based on
https://github.com/puppetlabs/puppetlabs-mysql/pull/1057 and depends on https://tickets.puppetlabs.com/browse/MODULES-6932

----

We should not merge this before the MODULES-6932 issue is solved

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -14,7 +14,7 @@
     },
     {
       "name": "puppetlabs/mysql",
-      "version_requirement": ">= 3.5.0 < 6.0.0"
+      "version_requirement": ">= 5.4.0 < 6.0.0"
     },
     {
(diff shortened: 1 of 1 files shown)


ekohl:
If I understand this correct we need pl-mysql >= 5.4.0 for ArchLinux compatibility? IMHO that's not a real reason to bump the minimum version in this module because API-wise we're still compatible with the older version.

Other than that the title doesn't really match the content :)

bastelfreak:
after some discussion we won't merge this. the API in the mysql version is correct and works with the module. this bug is arch specific but we already allow newer versions that work, so this is fine.

Issue #505, opened 2018-06-20
Exec[zabbix_server_create.sql] does not work because is might come after Package[zabbix-server-mysql


stefanandres:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.5.1
- Distribution: ubuntu-16.04
- Module version: 6.3.1

Puppet fails with
```
       Notice: /Stage[main]/Zabbix::Database::Mysql/Exec[zabbix_server_create.sql]/returns: /bin/sh: 1: cd: can't cd to /usr/share/doc/zabbix-*-mysql*
       Error: 'cd /usr/share/doc/zabbix-*-mysql* && if [ -f create.sql.gz ]; then gunzip -f create.sql.gz ; fi && mysql -h '127.0.0.1' -u 'zabbix_server' -p'test1234' -D 'zabbix_server' < create.sql && touch /etc/zabbix/.schema.done' returned 2 instead of one of [0]
       Error: /Stage[main]/Zabbix::Database::Mysql/Exec[zabbix_server_create.sql]/returns: change from 'notrun' to ['0'] failed: 'cd /usr/share/doc/zabbix-*-mysql* && if [ -f create.sql.gz ]; then gunzip -f create.sql.gz ; fi && mysql -h '127.0.0.1' -u 'zabbix_server' -p'test1234' -D 'zabbix_server' < create.sql && touch /etc/zabbix/.schema.done' returned 2 instead of one of [0]
``` 

because the package might be installed afterwards:
```
       Notice: /Stage[main]/Zabbix::Server/Package[zabbix-server-mysql]/ensure: created
```

https://github.com/voxpupuli/puppet-zabbix/blob/master/manifests/database/mysql.pp#L78 is missing requires for this package.

stefanandres:
Fixed with https://github.com/voxpupuli/puppet-zabbix/pull/508

PR #508, opened 2018-06-20
Add missing dependency for sql files on their Package

stefanandres:
See https://github.com/voxpupuli/puppet-zabbix/issues/505


bastelfreak:
Thanks for the bugfix @stefanandres !

bastelfreak:
@stefanandres can you take a look at the travis failures?

stefanandres:
Yes, of course :-)

stefanandres:
Ah, in fact I was using it wrong.

The workflow should be:
Use the following classes (in this Order):
* zabbix::database
* mysql::server
* zabbix::server

Before I was using zabbix::database::mysql directly, but database.pp already requires on the package. To what is to be done is to make zabbix::database::mysql a private class. I'll adjust the PR for this. 
This will help new users not to confuse it also.

Issue #521, opened 2018-08-07
PrivateTmp always enabled, this messes with SNMP trap configurations


ranzbak:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: All versions
- Ruby: na
- Distribution: Oracle Linux 7
- Module version: 5.0.0

## How to reproduce (e.g Puppet code you use)
Build a machine using the module, automatically systemd service file will contain : PrivateTmp=yes
I would like to be able to configure this to be PrivateTmp=yes

## What are you seeing
I'm seeing no way to configure this setting, leaving me unable to configure SNMP traps.

## What behaviour did you expect instead
I expected a Class argument to enable and disable the PrivateTmp systemd setting.
The default should still be 'yes', but 'no' should be a configurable option.

## Output log
na

## Any additional information you'd like to impart
In onder to have SNMP traps being passed to Zabbix a path has to be accessible to write a logfile to, Zabbix can read. https://zabbix.org/wiki/Start_with_SNMP_traps_in_Zabbix

ranzbak:
What also could be a solution is to add : JoinsNamespaceOf=
To the systemd file so both the feeder and Zabbix share the same tmp directory.
But this doesn't make debugging the setup easy, at least for debugging the private tmp should be configurable.

bastelfreak:
that sounds good, :+1: from my side.

PR #522, opened 2018-08-09
Add switch to disable privatetmp in the server systemd file.

ranzbak:
This pool request adds a switch to the server class, to enable/disable the privatetmp option in the zabbix-server systemd file.

PrivateTmp always enabled, this messes with SNMP trap configurations #521

### DIFF
--- a/.msync.yml
+++ b/.msync.yml
@@ -1,1 +1,1 @@
-modulesync_config_version: '1.9.4-rc0'
+modulesync_config_version: '1.9.6'
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,25 +28,25 @@ matrix:
   - rvm: 2.5.1
     bundler_args: --without development release
     dist: trusty
-    env: PUPPET_INSTALL_TYPE=agent BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_debug=true BEAKER_setfile=ubuntu1404-64{hypervisor=docker} CHECK=beaker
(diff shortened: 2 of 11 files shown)


bastelfreak:
Hi @ranzbak, thanks for the PR. I'm wondering if this is the right solution to the problem. This means the shared directory will be /tmp? I don't think that this is a good idea. Wouldn't it make more sense to create a new directory which is shared between both processes exclusively?

bastelfreak:
Please rebase after https://github.com/voxpupuli/puppet-zabbix/pull/524 got merged. That will fix the failing beaker jobs

ranzbak:
What also could be a solution is to add : JoinsNamespaceOf=
To the systemd file so both the feeder and Zabbix share the same tmp directory.
But this doesn't make debugging the setup easy, at least for debugging the private tmp should be configurable.

ranzbak:
Also rebase done :-)

bastelfreak:
@ranzbak the rebase doesn't look, can you give it another shot? If you need any help with git, feel free to join our slack channel at https://slack.puppet.com or our IRC channel #voxpupuli on freenode.

Fabian1976:
I also ran in to this issue. I fixed it by letting snmptt write the file in this location:
```/var/log/zabbix/zabbix_traps.tmp```
And setting the trapper file of the server process to the same location.

bastelfreak:
@Fabian1976 are you interested in finishing this PR?

Fabian1976:
@bastelfreak No, not really :)  I agree with you that this isn't the right solution to "fix" it. Just write the trapper file to another folder (not /tmp) where it can also be managed by logrotate and archived if you want to have an archive.

bastelfreak:
Hi. I'm going to close this PR. Please reopen it if you still think this is a good idea and we can discuss it here or on slack http://slack.puppet.com/ or our IRC channel #voxpupuli on freenode.

PR #530, opened 2018-08-26
allow puppetlabs/stdlib 5.x, puppetlabs/apt 6.x and puppetlabs/concat 5.x

bastelfreak:


bastelfreak:
postgres pulls in an old version of concat will pulls in stdlib 4.X. We need to wait for a new postgres release.

bastelfreak:
Requested a release at https://tickets.puppetlabs.com/browse/MODULES-7744

bastelfreak:
now fails because of mysql  not supporting latest stdlib :(

PR #545, opened 2018-09-27
Gentoo service fix

lordievader:
This PR continues the Gentoo support of the puppet-zabbix module.
It fixes a few forgotten changes from `zabbix-agent` to `$servicename`. Which is sourced from `params.pp`.

In the second commit a copy is made of `zabbix-agent-systemd.init.erb` and named as `zabbix-agentd-systemd.init.erb`. I'm not sure if this is the wanted procedure, hence I left it as a separate commit. If this is not the correct procedure feel free to reject it ;)

The last commit adds `$servicename` like support to userparameters.pp.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -346,6 +346,7 @@ class zabbix::agent (
     zabbix_user               => $zabbix_user,
     additional_service_params => $real_additional_service_params,
     service_type              => $real_service_type,
+    service_name              => 'zabbix-agent',
     require                   => Package[$zabbix_package_agent],
   }
 
@@ -362,7 +363,7 @@ class zabbix::agent (
   } else {
     $service_provider = undef
   }
-  service { 'zabbix-agent':
+  service { $servicename:
     ensure     => running,
     enable     => true,
     provider   => $service_provider,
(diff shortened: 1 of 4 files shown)


Dan33l:
Hey @lordievader , thank you for the PR.

Accordingly to our reviewing policy new template are prefered with `epp` and not with `erb`.
https://voxpupuli.org/docs/#reviewing-a-module-pr

And you need to fix tests, some of them failed.

baurmatt:
The file `templates/zabbix-agentd-systemd.init.erb` shouldn't be needed due it's content being exactly the same as `zabbix-agent-systemd.init.erb`

Dan33l:
> The file `templates/zabbix-agentd-systemd.init.erb` shouldn't be needed due it's content being exactly the same as `zabbix-agent-systemd.init.erb`

So can you modify your PR to remove its use and fix tests ?

lordievader:
@baurmatt but via commit 007b5272514c0806591290dedbba21415f4533cc the module tries to find a template named 'zabbix-agentd-systemd.init'. This is the reason why I duplicated the service file.

@Dan33l I will try to fix the tests when I have time. Have been quite busy with other things the past few weeks, sorry.

lordievader:
Oh, that does sound like a nicer fix. I'll implement that :)

bastelfreak:
@lordievader can you rebase against our master and take a look at the failing spec tests?

lordievader:
@bastelfreak After quite a few attemps, I fixed the failing tests. As Gentoo doesn't seem part of the unit tests (I didn't see an easy way of adding the OS), I tested that part on my own system.

bastelfreak:
Thanks!

PR #552, opened 2018-10-17
Fix proxy documentation

frenchtoasters:
I updated the zabbix-proxy section to account for the database type error message `Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Class[Zabbix::Database]: parameter 'database_type' expects a match for Zabbix::Databases = Enum['mysql', 'postgresql', 'sqlite']` when using the example documentation that was provided.


bastelfreak:
Thanks @frenchtoasters !

PR #553, opened 2018-10-22
Add manage_init_script parameter

### LINKED ISSUE #541: Drop management of Init script

- Puppet: All
- Ruby: All
- Distribution: All
- Module version: Since #240

## What are you seeing
The module manages the init script for agent and server. This was done because some custom settings e.g. `configdir` require a change for startup values of the zabbix agent|server process. In my opinion managing the init file is bad because it leads to a lot of problems with e.g. updates and should only be done if there is *no* other option.

Use the OS's `defaults` file:

Debian -> /etc/default/zabbix-agent
RedHat -> /etc/sysconfig/zabbix-agent
SystemD -> /etc/systemd/system/zabbix-agent.service.d/override.conf

Are you interested (and are willing to merge it ;)) in a PR which moves the configuration to the mentioned default files?

baurmatt:
Adds a parameter `manage_init_script` which makes it possible to disable the Init File management.

Fixes #541.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -177,6 +177,9 @@
 # [*loadmodule*]
 #   Module to load at agent startup.
 #
+# [*manage_startup_script*]
+#  If the init script should be managed by this module. Attention: This might cause problems with some config options of this module (e.g agent_configfile_path)
+#
 # === Example
 #
 #  Basic installation:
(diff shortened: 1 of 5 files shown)


baurmatt:
I'm pretty sure the failing test isn't related to my code change. In fact, i can't even reproduce this on my local machine oO

manifests/agent.pp
```diff
@@ -177,6 +177,9 @@
 # [*loadmodule*]
 #   Module to load at agent startup.
 #
+# [*manage_init_script*]
```
alexjfisher: `zabbix::startup` manages systemd instead of init scripts on some systems, so this parameter name may not be the best choice.

alexjfisher: works for me.  @bastelfreak - sound ok?

bastelfreak: sounds good to me. @baurmatt can you update it please?

baurmatt: Done. Didn't change the branch name though ;)

bastelfreak:
@alexjfisher can you have a look again?

PR #555, opened 2018-10-26
Replace is_ip_address with Puppet 4 native comparision

baurmatt:

### DIFF
--- a/manifests/proxy.pp
+++ b/manifests/proxy.pp
@@ -429,7 +429,7 @@ class zabbix::proxy (
   if ($listenip != undef) {
     if ($listenip =~ /^(eth|lo|bond|lxc|eno|tap|tun).*/) {
       $listen_ip = getvar("::ipaddress_${listenip}")
-    } elsif is_ip_address($listenip) {
+    } elsif $listenip =~ Stdlib::IP::Address {
       $listen_ip = $listenip
     } else {
       $listen_ip = undef
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
     },
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.19.0 < 6.0.0"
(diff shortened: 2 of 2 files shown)


baurmatt:
Same error as in #553. I'm pretty confident that this has nothing to do with this change.

manifests/proxy.pp
```diff
@@ -429,7 +429,7 @@
   if ($listenip != undef) {
     if ($listenip =~ /^(eth|lo|bond|lxc|eno|tap|tun).*/) {
       $listen_ip = getvar("::ipaddress_${listenip}")
-    } elsif is_ip_address($listenip) {
+    } elsif $listenip =~ Stdlib::Compat::Ip_address {
```
alexjfisher: Should this use `Stdlib::IP::Address` instead of `Stdlib::Compat::Ip_address`?

Dan33l: `metadata.json` requires `stdlib >= 4.19.0 < 6.0.0`.
Data type `Stdlib::IP::Address` requires  at least `stdlib 4.25.0`.

baurmatt: @Dan33l That's the reason why I chose `Stlib::Compat::Ip_address` :) If its ok to increase the stdlib dependency to 4.25.0, I'm happy to use `Stdlib::IP::Address`.

alexjfisher: I'd be fine with increasing the stdlib dependency.

alexjfisher:
@baurmatt Test failure is fixed in https://github.com/voxpupuli/puppet-zabbix/pull/557 (which should be merged shortly)

baurmatt:
@alexjfisher Thanks, rebased from master! :)

alexjfisher:
@baurmatt Thanks.  I had actually already restarted your build, (and when travis builds a PR, it builds the candidate merge commit)

PR #559, opened 2018-11-02
Improve performance of zabbix_host

### LINKED ISSUE #558: Improve performance of zabbix_host

- Puppet: Puppet 5.5.6
- Ruby: Ruby 2.4.3
- Distribution: Ubuntu 16.04
- Module version: Current master

(The hosts are already created in Zabbix)

```puppet
$range = range(0, 3000)

$range.each |$i| {
  zabbix_host { "test${i}":
    templates    => ['Template OS Linux'],
    ipaddress    => '127.0.0.1',
    port         => 10050,
    use_ip       => true,
    group        => 'testhosts',
    group_create => true,
    zabbix_url   => 'zabbix-host.example.de',
    zabbix_user  => 'Admin',
    zabbix_pass  => 'zabbix',
  }
}
```

```bash
root@mbaur-zabbix:~# puppet apply test.pp
Notice: Compiled catalog for mbaur-zabbix.restvlan6.syseleven in environment production in 2.20 seconds
Notice: Applied catalog in 126.52 seconds
```

## What are you seeing
The performance of zabbix_host is really poor. The problems seems to be, that Puppet does, for the a single host as definied above, 7 API calls. If there would be an additional template the API call count goes up to 13 and increases even further for more templates.

Puppet does 2 API calls:
* login
* get host with templates

baurmatt:
Before this commit, every template added to a host genarated addtional 5
API calls. This isn't needed as we can get all required information by
only one API call.

Test with 3000 hosts:
```
root@mbaur-zabbix:~# puppet apply test.pp
Notice: Compiled catalog for mbaur-zabbix.restvlan6.syseleven in environment production in 2.21 seconds
Notice: Applied catalog in 49.49 seconds
```

This is around 76s fast then before.

Fixes #558.


bastelfreak:
Thanks for the PR!

PR #378, opened 2017-05-02
Add ldap reqcert

elmobp:
Add the option to specify the ldap_reqcert this is needed in some SSL implementations

PR #560, opened 2018-11-02
Add ldap reqcert

bastelfreak:
This is a resubmit of https://github.com/voxpupuli/puppet-zabbix/pull/378

----

Add the option to specify the ldap_reqcert this is needed in some SSL implementations


manifests/web.pp
```diff
@@ -151,6 +151,9 @@
 # [*ldap_clientkey*]
 # Set location of client key used by LDAP authentication.
 #
+# [*ldap_reqcert *]
+# Specifies what checks to perform on a server certificate ( never | allow | try | demand,hard )
```
ekohl: The `demand,hard` is a bit odd. However, since there's a data type I'd suggest to leave them out. If you generate a `REFERENCE.md` the data type should be enough.

bastelfreak: updated it. I checked the docs, `demand` and `hard` are independent values.

PR #562, opened 2018-11-03
Add ability to stop the agent

mkilchhofer:
For our "IoT use case" we want to completely configure a mini server during the staging process. When a technician installs the device to its final position (in a vehicle), we want to activate the monitoring agent.

This PR adds the ability to start/stop the agent and to enable/disable the agents automatic startup during boot.

--> Could someone help me writing my first rspec tests for this feature?

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -237,7 +237,7 @@ class zabbix::agent (
   $listenip                               = $zabbix::params::agent_listenip,
   $startagents                            = $zabbix::params::agent_startagents,
   $serveractive                           = $zabbix::params::agent_serveractive,
-  Enum['running','stopped'] $service_state  = $zabbix::params::agent_service_state,
+  Stdlib::Ensure::Service $service_state  = $zabbix::params::agent_service_state,
   Boolean $service_enable                 = $zabbix::params::agent_service_enable,
   $hostname                               = $zabbix::params::agent_hostname,
   $hostnameitem                           = $zabbix::params::agent_hostnameitem,
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -94,6 +94,15 @@
 # [*serveractive*]
 #   List of comma delimited ip:port (or hostname:port) pairs of zabbix servers for active checks.
 #
+# [*service_state*]
+#   Start / stop the agent service. E.g. to preconfigure a hosts agent and turn on the service
+#   at a later time (when the server reaches production SLA)
+#   Default: 'running'
+#
+# [*service_enable*]
+#   Automatically start the agent on system boot
+#   Default: true
+#
 # [*hostname*]
 #   Unique, case sensitive hostname.
 #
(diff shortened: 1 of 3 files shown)


manifests/agent.pp
```diff
@@ -228,6 +237,8 @@
   $listenip                               = $zabbix::params::agent_listenip,
   $startagents                            = $zabbix::params::agent_startagents,
   $serveractive                           = $zabbix::params::agent_serveractive,
+  Enum['running','stopped'] $service_state  = $zabbix::params::agent_service_state,
```
bastelfreak: Can you instead use the stdlib datatype for this:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/ensure/service.pp#L1

bastelfreak:
Hi @mkilchhofer, thanks for the contribution! Did you already check https://github.com/voxpupuli/puppet-zabbix/blob/master/spec/classes/agent_spec.rb? That is the file with all the related rspec-puppet tests for the agent. If you've any questions I suggest you join our slack channel at https://slack.puppet.com or our IRC channel #voxppuli on freenode. That would make it way easier to discuss proper testing and an rspec walkthrough.

manifests/params.pp
```diff
@@ -239,6 +239,8 @@
   $agent_refreshactivechecks                = '120'
   $agent_server                             = '127.0.0.1'
   $agent_serveractive                       = undef
+  $agent_service_ensure                      = 'running'
```
bastelfreak: can you fix the `=` alignment here?

mkilchhofer: yup :)

mkilchhofer:
> Hi @mkilchhofer, thanks for the contribution! Did you already check https://github.com/voxpupuli/puppet-zabbix/blob/master/spec/classes/agent_spec.rb? That is the file with all the related rspec-puppet tests for the agent. If you've any questions I suggest you join our slack channel at https://slack.puppet.com or our IRC channel #voxppuli on freenode. That would make it way easier to discuss proper testing and an rspec walkthrough.

I will have a look into rspec tests in the next week. I had my problems to run them on the rhel machines at work (ruby2.0). But on my fedora laptop (ruby2.5+), your mentioned spec file works fine.

bastelfreak:
We don't support ruby2.0 and 2.1 anymore. Both are EOL since ages. Ruby2.2 and newer should work.

mkilchhofer:
Is something like this enough:

```
diff --git a/spec/classes/agent_spec.rb b/spec/classes/agent_spec.rb
index da00269..0be24d7 100644
--- a/spec/classes/agent_spec.rb
+++ b/spec/classes/agent_spec.rb
@@ -243,6 +243,25 @@ describe 'zabbix::agent' do
 
         it { is_expected.to contain_file(config_path).without_content %r{^ListenIP=} }
       end
+
+      context 'when declaring service_ensure is stopped and service_enable false' do
+        package = 'zabbix-agent'
+
+        let :params do
+          {
+            service_ensure: 'stopped',
+            service_enable: false
+          }
+        end
+
+        it do
+          is_expected.to contain_service('zabbix-agent').with(
+            ensure:     'stopped',
+            enable:     false,
+            require:    "Package[#{package}]"
+          )
+        end
+      end
     end
   end
 end
```

bastelfreak:
@mkilchhofer yes the test looks fine. I restarted the failed travis job.

PR #572, opened 2018-12-12
Properly require zabbixapi gem

baurmatt:
This PR changes how we `require` the `zabbixapi` to the recommended way by creating a feature for it which we later confine.

-


baurmatt:
I have no idea why the test breaks... If I comment out `confine feature: :zabbixapi` in `lib/puppet/provider/zabbix_template/ruby.rb` it works fine.

@alexjfisher Do you have an idea why? git history shows that you implemented the test.

baurmatt:
So seems like not getting a good night of sleeps helps solving problems.... The problem seems to be, that the `zabbixapi` gem isn't available in the tests. After adding that, the tests run fine.

alexjfisher:
I can't work out how it passed before though...

PR #576, opened 2018-12-17
Fix rspec tests for Gentoo

baurmatt:
Due to the release of facterdb 0.6.0 we're now testing against Gentoo as
well. This leads to a couple of errors which this commit fixes.

~Strict variable checking was disabled until
https://github.com/puppetlabs/puppetlabs-mysql/pull/1147 got merged.~ --> Merged.


baurmatt:
Travis seems to have problems :( There is always one Beaker test failing :/

alexjfisher:
@baurmatt Thanks!

Issue #570, opened 2018-12-12
Move zabbix_(url|user|pass) and apache_use_ssl to config file


baurmatt:
(I removed the default template because doesn't really apply to a feature)

### Problem to solve

The current zabbix types/provider are pretty hacky. Some examples:
* `puppet resource` purging isn't possible (`self.instances` isn't implemented)
* `zabbix_host` doesn't allow `ipaddress`, `use_ip`, `proxy`, ... to be changed
* If templates get added to/from `zabbix_host` the change isn't properly show. Puppet only says `Notice: /Stage[main]/Zabbix_host[example.org]/ensure: current_value 'absent', should be 'present'`
* Other types/provider have most certainly the same problem as `zabbix_host`

### Proposal
* Implement `self.instances`/`self.prefetch` for all providers and change current type params to properties where needed. This will allow Puppet itself to determine which things need to change and will allow proper logging of the change `Notice: /Stage[main]/Main/Zabbix_host[example.org]/ipaddress: current_value '10.0.0.2', should be '10.0.0.1'`
* To implement `self.instances` we need the credentials to be available outside of the type. This is required because `self.instances` cannot access the `zabbix_(url|user|pass)` params of the custom type

bastelfreak:
Hi @baurmatt, the proposal sounds good. Sadly I won't have time in the next weeks to implement something like this. If you're willing to do this I'm happy to give it a review and merge it.

baurmatt:
Hey @bastelfreak, thanks for taking the time to review my proposal. I'm happy to implement this myself, in fact I'm already have a large portion of it implement in our fork. I will add PRs over time which inclemently implement this.

PR #579, opened 2018-12-21
Move api configuration to a config file

baurmatt:
This is the first change for enhancing the custom types with deeper Puppet
integration. This is needed to ensure the implementation of
self.instances/self.prefetch works correctly.

This is the basis for all changes needed for #570.


bastelfreak:
This looks good to me, but I would like to see another review. @ekohl or @dhollinger maybe?

bastelfreak:
@baurmatt can you please rebase?

baurmatt:
Done :)

ekohl:
I think that the types should autorequire the config file. Not sure if that means the collectors could be replaced just yet. A host might require a template, but perhaps that can also be solved via autorequires.

bastelfreak:
@ekohl I am fine with this. What do you think?

baurmatt:
Hey, anything left we need to talk about? :) Would love to see this merged!

manifests/resources/web.pp
```diff
@@ -13,38 +13,32 @@
...
-    zabbix_pass    => $zabbix_pass,
-    apache_use_ssl => $apache_use_ssl,
-  }
-  -> Zabbix_userparameters <<| |>> {
-    zabbix_url     => $zabbix_url,
-    zabbix_user    => $zabbix_user,
-    zabbix_pass    => $zabbix_pass,
-    apache_use_ssl => $apache_use_ssl,
-  }
+  -> Zabbix_template <<| |>>
```
alexjfisher: Are these needed anymore?  Perhaps they are redundant (given the types all autorequire the API config file)?

It depends on whether they were just being used to set the API credentials or whether actually eg. all Zabbix_userparameters have to come after all Zabbix_hosts etc?

> When given a block of attributes and values, a collector will set and override those attributes for every resource (virtual or not) that matches its search expression.
https://puppet.com/docs/puppet/5.3/lang_resources_advanced.html#amending-attributes-with-a-collector

baurmatt: From my understanding they ensure that e.g. the host exists in Puppet before the template gets added to it.

alexjfisher: Fair enough.  Just checking that the relationships before this change were actually all required (and not just a slightly lazy syntax short-cut).  Now they definitely look like they're there for a reason, so it's good to check.

baurmatt:
Still looking to get this merged ;) Is there anything you guys don't feel comfortable with?

alexjfisher:
@baurmatt Thanks!

PR #593, opened 2019-02-22
Implement self.prefetch for zabbix_hostgroup

baurmatt:
This is another change for #570.

### DIFF
--- a/lib/puppet/provider/zabbix_hostgroup/ruby.rb
+++ b/lib/puppet/provider/zabbix_hostgroup/ruby.rb
@@ -2,6 +2,24 @@ require_relative '../zabbix'
 Puppet::Type.type(:zabbix_hostgroup).provide(:ruby, parent: Puppet::Provider::Zabbix) do
   confine feature: :zabbixapi
 
+  def self.instances
+    api_hostgroups = zbx.hostgroups.all
+    api_hostgroups.map do |group_name, _id|
+      new(
+        ensure: :present,
+        name: group_name
+      )
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/zabbix_hostgroup/ruby.rb
```diff
@@ -2,14 +2,32 @@
...
+      if (resource = resources[prov.name])
+        resource.provider = prov
+      end
+    end
+  end
+
   def create
     # Connect to zabbix api
     hgid = zbx.hostgroups.create(name: @resource[:name])
     hgid
```
alexjfisher: Is there any reason for assigning to `hgid` here?  `create` doesn’t actually need to return anything, does it?

lib/puppet/provider/zabbix_hostgroup/ruby.rb
```diff
@@ -2,14 +2,32 @@
 Puppet::Type.type(:zabbix_hostgroup).provide(:ruby, parent: Puppet::Provider::Zabbix) do
   confine feature: :zabbixapi
 
+  def self.instances
+    api_hostgroups = zbx.hostgroups.all
+    api_hostgroups.map do |group_name, _id|
+      new(
+        ensure: :present,
+        name: group_name
```
alexjfisher: Is `provider: :ruby` needed here?
Or is that just what I’ve done in the past by blindly following what I read somewhere once?! (Probably http://garylarizza.com/blog/2013/12/15/seriously-what-is-this-provider-doing/)

PR #599, opened 2019-04-02
Put selboolean{'zabbix_can_network'} inside ensure_resources

### LINKED ISSUE #598: Cannot declare Selboolean zabbix_can_network elsewhere

- Puppet: 5.5.10 AIO
- Ruby: shipped with AIO
- Distribution: CentOS 7
- Module version: 6.7.0

Use sth like `site.pp` to globally declare selboolean { 'zabbix_can_network' } and on specified node declare zabbix::server.

## What are you seeing
Puppet fails because there is duplicate resource declaration (caused by selboolean)
I'd like to safely duplicate that resource declaration

I'll provide PR fixing this issue

KrzysztofHajdamowicz:
puppet-zabbix module depends on stdlib, `ensure_resources` is already availble for usage.
My motivation for this change is global declaration that zabbix_can_network and zabbix server belongs to this global declaration.
ensure_resource allows for resource duplication as long as declaration of those resource is identical.
Also, notifying service dependency is useless IMO, as SELinux and ZBX are separate things and ZBX doesn't need to be restarted in order to make SELinux policies to be applied.

Fixes #598

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -567,15 +567,18 @@ class zabbix::server (
   }
   # check if selinux is active and allow zabbix
   if $facts['selinux'] == true and $manage_selinux {
-    selboolean{'zabbix_can_network':
-      persistent => true,
-      value      => 'on',
-      notify     => $dependency,
-    }
-    -> selinux::module{'zabbix-server':
+    ensure_resource ('selboolean',
+      [
+        'zabbix_can_network',
+      ], {
+        persistent => true,
+        value      => 'on',
+      })
+    selinux::module{'zabbix-server':
       ensure    => 'present',
       source_te => 'puppet:///modules/zabbix/zabbix-server.te',
       before    => $dependency,
+      require   => Selboolean['zabbix_can_network']
     }
     # zabbix-server 3.4 introduced IPC via a socket in /tmp
     # https://support.zabbix.com/browse/ZBX-12567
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi. Can you please expain why you wrap the selboolean into `ensure_resource()`? The `zabbix_can_network` setting is pretty specific to selinux so in my opinion should be managed within this module.

KrzysztofHajdamowicz:
> Hi. Can you please expain why you wrap the selboolean into `ensure_resource()`? The `zabbix_can_network` setting is pretty specific to selinux so in my opinion should be managed within this module.

`zabbix_can_network` seboolean allows both zabbix-agent and zabbix-server to create TCP connections.
In our usecase we want do check did zbxagnt can connect to specified TCP port (service stops listening at it, despite it is still running so systemd can't complain). Such service exists on all hosts, including Zabbix Server, thus duplicate resource declaration.

`ensure_resource` looks like good solution, because, according to [documentation](https://github.com/puppetlabs/puppetlabs-stdlib)
> If the resource already exists, but does not match the specified parameters, this function attempts to recreate the resource, leading to a duplicate resource definition error

So puppet will run with success ONLY IF all declarations are the same. Thus, no difference if it resource is managed by puppet-zabbix or elsewhere. This resource can be only in state puppet-zabbix wants or puppet run will fail.

KrzysztofHajdamowicz:
Also, can You retrigger unit test? All failed ones are due to timeout od VM with docker.

bastelfreak:
I restarted all failed jobs.

KrzysztofHajdamowicz:
Looks like unittests are OK :)

PR #623, opened 2019-09-19
Upgrade zabbixapi package for Zabbix 4.2 support

Bouke:
zabbixapi 4.0.0 doesn't support Zabbix 4.2.

### DIFF
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -278,7 +278,7 @@ class zabbix::web (
         $zabbixapi_version = '3.2.1'
       }
       default : {
-        $zabbixapi_version = '4.1.1'
+        $zabbixapi_version = '4.1.2'
       }
     }
 
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -278,7 +278,7 @@ class zabbix::web (
         $zabbixapi_version = '3.2.1'
       }
       default : {
-        $zabbixapi_version = 'latest'
+        $zabbixapi_version = '4.1.2'
       }
     }
 
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -278,7 +278,7 @@ class zabbix::web (
         $zabbixapi_version = '3.2.1'
       }
       default : {
-        $zabbixapi_version = '4.1.2'
+        $zabbixapi_version = 'latest'
       }
     }
 
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -278,7 +278,7 @@ class zabbix::web (
         $zabbixapi_version = '3.2.1'
       }
       default : {
-        $zabbixapi_version = '4.0.0'
+        $zabbixapi_version = '4.1.1'
       }
     }
 


bastelfreak:
Hi, thanks for the PR. Can you please rebase after https://github.com/voxpupuli/puppet-zabbix/pull/624 got merged?

Bouke:
I think you can merge the PR and have GitHub rebase it?

bastelfreak:
@Bouke yes that works. But that won't kick of travis before the merge. Please rebase. Afterwards travis will test your changes.

Bouke:
I've rebased the PR.

bastelfreak:
@Bouke can you please have a look at the failing travis tests?

Bouke:
It looks like the new API version has an unsolved issue which the tests run into https://github.com/express42/zabbixapi/issues/99. I think the only solution here is to wait until that bug is resolved in a future release.

Bouke:
If anybody knows how to deploy a fork hosted on github using puppet_gem package provider, it would be very helpful.

Bouke:
With the following in place, everything seems to work fine on Zabbix 4.4. It just logs a lot of "unsupported messages", but new resources seem to be added just fine.
```
    Package<| title == 'zabbixapi' |> {
        ensure => 'latest',
    }
```

baurmatt:
@Bouke I would prefer not to add 'latest' to packages because it will update the package as soon as a new release gets out. Can we just set the variable to "4.1.2"? That's the latest release currently available on Rubygems.

Bouke:
> @Bouke I would prefer not to add 'latest' to packages because it will update the package as soon as a new release gets out. Can we just set the variable to "4.1.2"? That's the latest release currently available on Rubygems.

Ideally as a consumer of this module, I'd like the ability to override such versions. Currently it just hardcodes some version (either 'latest' or '4.1.2'). Maybe we can change the parameters to allow overriding the version?

bastelfreak:
I agree that this should be a parameter. But we also need to provide sane default that work with up2date zabbix versions. I did a lot of fiddling around in https://github.com/voxpupuli/puppet-zabbix/pull/684/files to provide matching gem versions, but I failed hard (please ignore the labels).

Bouke:
I've rebased the PR and made the change to hardcode version 4.1.2. So should be all good now.

bastelfreak:
I now see the following message in the acceptance test output:
```
  [WARN] Zabbix API version: 3.0.31 is not supported by this version of zabbixapi
```


also we've a few places where we hardcode the zabbix version in the test to 3 because of old limitations from the gem:
https://github.com/voxpupuli/puppet-zabbix/blob/master/spec/acceptance/zabbix_template_spec.rb#L17
I think we should try to test all the supported zabbix versions to ensure we choose the correct zabbixapi gem version?

Bouke:
I agree that it would be good for this package to have good test coverage against recent releases of the various tools it configures. Meanwhile I would also suggest to leave that out of the scope of this PR.

bastelfreak:
@Bouke can you please rebase against our latest master branch? I still think that the zabbix versions in our acceptance tests should be updated in this PR. The PR claims to support zabbix 4.2, but we only test on 3.0.

bastelfreak:
thanks for all the work everybody. we merged the changes in https://github.com/voxpupuli/puppet-zabbix/pull/730

PR #627, opened 2019-10-06
add puppet-lint-legacy_facts-check gem & puppet-lint: fix legacy_facts

bastelfreak:


bastelfreak:
This is now already in master

PR #628, opened 2019-10-06
drop legacy puppetlabs/pe_gem dependency & cleanup code

bastelfreak:
This module was required to support EOL Puppet/Puppet Enterprise
versions. We don't support them since a long time, so we can remove the
dependency + related code.


bastelfreak:
ups. I had it already removed from the metadata.json but it got readded during a rebase. I don't think that this is a breaking change since we officially dropped support for those puppet/pe versions some time ago in another release. But since the next release will be a major one we could also mark it as `backwards-incompatible`, just to be safe.

PR #629, opened 2019-10-06
release v7.0.0

bastelfreak:


bastelfreak:
I checked a few of the changes in the older releases. The old entries were wrong and the new ones are correct.

PR #634, opened 2019-10-19
drop BEAKER_IS_PE=no beaker argument

bastelfreak:


bastelfreak:
this is now in master

PR #637, opened 2019-10-25
Add support for adding a JMX interface to a host

Fabian1976:
Add support for adding a JMX interface to host

Just add this to your hiera:
```
zabbix::agent::jmx_port: 9999
```

### DIFF
--- a/lib/puppet/provider/zabbix_host/ruby.rb
+++ b/lib/puppet/provider/zabbix_host/ruby.rb
@@ -51,19 +51,34 @@ Puppet::Type.type(:zabbix_host).provide(:ruby, parent: Puppet::Provider::Zabbix)
     proxy_hostid = @resource[:proxy].nil? || @resource[:proxy].empty? ? nil : zbx.proxies.get_id(host: @resource[:proxy])
 
     # Now we create the host
-    zbx.hosts.create(
-      host: @resource[:hostname],
-      proxy_hostid: proxy_hostid,
-      interfaces: [
+    default_interface = [
+      {
+        type: 1,
+        main: 1,
+        ip: @resource[:ipaddress],
+        dns: @resource[:hostname],
+        port: @resource[:port],
+        useip: @resource[:use_ip] ? 1 : 0
+      }
+    ]
+    if jmx_port != 0 and !jmx_port.nil?
+      host_interfaces = default_interface + [
         {
-          type: 1,
+          type: 4,
           main: 1,
           ip: @resource[:ipaddress],
           dns: @resource[:hostname],
-          port: @resource[:port],
+          port: @resource[:jmx_port],
           useip: @resource[:use_ip] ? 1 : 0
         }
-      ],
+      ]
+    else
+      host_interfaces = default_interface
+    end
+    zbx.hosts.create(
+      host: @resource[:hostname],
+      proxy_hostid: proxy_hostid,
+      interfaces: host_interfaces,
       templates: templates,
       groups: groups
     )
--- a/lib/puppet/provider/zabbix_host/ruby.rb
+++ b/lib/puppet/provider/zabbix_host/ruby.rb
@@ -61,20 +61,20 @@ Puppet::Type.type(:zabbix_host).provide(:ruby, parent: Puppet::Provider::Zabbix)
         useip: @resource[:use_ip] ? 1 : 0
       }
     ]
-    if jmx_port != 0 and !jmx_port.nil?
-      host_interfaces = default_interface + [
-        {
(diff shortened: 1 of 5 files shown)


manifests/agent.pp
```diff
@@ -282,6 +285,7 @@
   String $additional_service_params               = $zabbix::params::additional_service_params,
   String $service_type                            = $zabbix::params::service_type,
   Boolean $manage_startup_script                  = $zabbix::params::manage_startup_script,
+  Integer $jmx_port                               = $zabbix::params::agent_jmx_port,
```
bastelfreak: Can you use the [Stdlib::Port](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/REFERENCE.md#stdlibport) datatype?
```suggestion
  Stdlib::Port $jmx_port                          = $zabbix::params::agent_jmx_port,
```

manifests/params.pp
```diff
@@ -278,6 +278,7 @@
     $agent_logfile                          = '/var/log/zabbix/zabbix_agentd.log'
     $agent_logfilesize                      = '100'
   }
+  $agent_jmx_port                           = 0
```
bastelfreak: Should the default be undef? In that case, the datatype should be `Optional[Stdlib::Port]`

Fabian1976: I use the check on 0 later on in the code. So it should be 0

Fabian1976:
What is the significance of the failed jobs? The puppet jobs seem to work. When will this be merged? Or is there something expected of me?

lib/puppet/type/zabbix_host.rb
```diff
@@ -101,6 +101,10 @@ def insync?(is)
     desc 'Whether it is monitored by an proxy or not.'
   end
 
+  newproperty(:jmx_port) do
```
baurmatt: Please add a data type validation.

lib/puppet/provider/zabbix_host/ruby.rb
```diff
@@ -51,19 +51,34 @@ def create
...
+      {
+        type: 1,
+        main: 1,
+        ip: @resource[:ipaddress],
+        dns: @resource[:hostname],
+        port: @resource[:port],
+        useip: @resource[:use_ip] ? 1 : 0
+      }
+    ]
+    host_interfaces = if jmx_port != 0 && !jmx_port.nil?
```
baurmatt: Did you check if this works with `self.instances`?

manifests/params.pp
baurmatt: Please check for nil/undef. Checking for 0 seems kind like a hack :)

baurmatt:
Hey @Fabian1976, thanks for this PR! :) I've left a couple of comments in the code, please address them. Regarding the failing tests: They look like they are related to your change. Could you please verify if your change actually works?

```
  Notice: Compiled catalog for ubuntu1404-64-1.c.travis-ci-prod-3.internal in environment production in 0.05 seconds
  Info: Applying configuration version '1572331677'
  Error: Could not set 'present' on ensure: Server answer API error
   {
    "code": -32602,
    "message": "Invalid params.",
    "data": "Incorrect arguments passed to function."
  }
   on request:
   {
    "method": "host.create",
    "params": [
      {
        "host": "test1.example.com",
        "interfaces": [
          {
            "type": 1,
            "main": 1,
            "ip": "127.0.0.1",
            "dns": "test1.example.com",
            "port": 10050,
            "useip": 1
          },
          {
            "type": 4,
            "main": 1,
            "ip": "127.0.0.1",
            "dns": "test1.example.com",
            "port": null,
            "useip": 1
          }
        ],
        "status": 0,
        "available": 1,
        "groups": [
          {
            "groupid": 8
          }
        ],
        "proxy_hostid": null,
        "templates": [
          {
            "templateid": 10001
          }
        ]
      }
    ],
    "id": 49721,
    "jsonrpc": "2.0",
    "auth": "0251d2407f003ca6b7bedfb02601f32f"
  } (file: /tmp/apply_manifest.pp.WYYzze, line: 1)
```

Fabian1976:
mmm, the problem is. I wrote it for version 6.7.0 and tested it and it worked. Then i saw version 7.0.0 had a lot of major changes in the inner workings so i tried to rewrite it without testing. I will close my PR and submit it again once we've upgrade to puppet 6 (still on puppet 4) and can use version 7.0.0.

fraenki:
AFAICT, the module still does not support adding JMX interfaces to a host. Is anybody else working on this? Otherwise I may try to resurrect the code from this PR...

PR #639, opened 2019-11-12
drop Ubuntu 14.04 support

bastelfreak:


baurmatt:
Failing tests are due to #645. Will be fixed with #646.

bastelfreak:
tahnks for the fix. I rebased this one and will take a look at the other failures later on.

PR #642, opened 2019-12-10
Implement self.prefetch for zabbix_proxy

baurmatt:
The template parameter got removed because it had no function and just added additional API calls. I therefore labeled this PR `backwards-incompatible`.

This is another change for #570.


baurmatt:
The tests which failed did this because of:

```
The job exceeded the maximum log length, and has been terminated.
```

Pretty sure that this isn't a problem of this change ;)

alexjfisher:
@baurmatt And not because https://github.com/voxpupuli/puppet-zabbix/pull/642/files#diff-677c6e628f74bf4e5fe32149fb37a725R29 printed a crazy amount of stuff to stdout?

baurmatt:
@alexjfisher LOL (literally) :D Good catch, thanks! :) I've updated the code. But there still two failing tests. As those are also failing in the master, I guess I'm good here.

baurmatt:
Tests will be fixed with #646.

PR #644, opened 2020-01-07
Add tablespace parameter

dpavlotzky:


manifests/params.pp
```diff
@@ -159,6 +159,7 @@
   $server_database_schema                   = undef
   $server_database_socket                   = undef
   $server_database_user                     = 'zabbix_server'
+  $server_database_tablespace               = 'pg_default'
```
baurmatt: Can we default this to `undef`? That's the default of the `tablespace` parameter of `postgresql::server::db`:

https://github.com/puppetlabs/puppetlabs-postgresql/blob/337595646688eec3c4f26ace51d1b3e8b7f5ad07/manifests/server/db.pp#L22

manifests/database.pp
```diff
@@ -72,6 +72,8 @@
 #   The default collation of the database.
 #   default: utf8_general_ci
 #
+# [* database_tablespace*]
```
baurmatt: Can you please clarify which database type this affects? It would be also nice to have the code failing if the parameter would be set for an other database type as postgresql

baurmatt:
Hey @dpavlotzky, thanks for you contribution! :) Can you please check the following things?

* Please address my comments
* Please add at least rspec tests for the change
* Please rebase the the branch after #646 got merged. It fixed the error with the failing CentOS 7 acceptance tests.

manifests/database.pp
```diff
@@ -129,6 +131,7 @@
   $database_host_ip                = $zabbix::params::server_database_host_ip,
   $database_charset                = $zabbix::params::server_database_charset,
   $database_collate                = $zabbix::params::server_database_collate,
+  $database_tablespace             = $zabbix::params::server_database_tablespace,
```
bastelfreak: Can you please add a datatype here? Probably a string:
```suggestion
  Optional[String[1]] $database_tablespace = $zabbix::params::server_database_tablespace,
```

manifests/params.pp
dpavlotzky: Done!

manifests/database.pp
dpavlotzky: Done!

manifests/database.pp
dpavlotzky: I fail to see why: the parameter always gets a default value from the params class.

baurmatt: Because we want database for all classes which aren't private :)

dpavlotzky: ?

alexjfisher: I think that should have read...

> We want data *types* for all classes which aren't private.

New parameters should be given types, even if there are other parameters that don't yet have any defined.

dpavlotzky: I can see how `String` would fit, but not `Optional`, since this parameter always has a value (`$zabbix::params::server_database_tablespace`) and will never cause a failure when the parameter is omitted.

manifests/database.pp
```diff
@@ -129,20 +132,27 @@
...
 ) inherits zabbix::params {
 
   # So lets create the databases and load all files. This can only be
   # happen when manage_database is set to true (Default).
   if $manage_database == true {
+    # Complain if database_tablespace is set and the database_type is not postgresql
+    if ($database_tablespace and $database_type != 'postgresql') {
+      notify { "database_tablespace is set to '${database_tablespace}'. This setting is only useful for PostgreSQL databases.": }
```
alexjfisher: I'd probably just go with a `fail` here instead of a notify resource.

dpavlotzky: IMHO a stale configuration option is not enough reason to have the catalog compilation fail, but `fail` is fine with me. Your call...

manifests/database.pp
dpavlotzky: > I think that should have read...
> 
> > We want data _types_ for all classes which aren't private.
> 
> New parameters should be given types, even if there are other parameters that don't yet have any defined.

👍

manifests/database.pp
alexjfisher: As notify isn't idempotent, I personally only tend to use it for cases when something will be fixed in the next puppet run.  eg maybe some puppet needs a fact before it can be part of your catalog, but that fact won't be present until after the first run.

```puppet
include java

if $facts['jdk_version'] {
  # Install something that needs java and needs to know what version.
  # ...
} else {
  notify {'jdk_version fact was missing. Foo was not installed.  Please run puppet agent to complete the configuration.':}
}
```

In this case, I think I'd prefer a hard fail rather than a notify resource reporting a 'change' on each run.  (But I don't feel that strongly either)

dpavlotzky: I asked my colleagues about this and they would use a fail as well, so a fail it will be.

baurmatt:
Hey @dpavlotzky, thanks for the changes! :) Could you please squash you commits so that I can cleanly merge it?

dpavlotzky:
Sorry! Messed up the squash :-(. Created a new pull-request #649

Issue #645, opened 2020-01-07
Test for CentOS 7 current failing due to bug in docker


baurmatt:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: All
- Ruby: All
- Distribution: CentOS 7.7
- Module version: Current master

## How to reproduce (e.g Puppet code you use)
Take a look at the failing Travis test in master (https://travis-ci.org/voxpupuli/puppet-zabbix/jobs/633797254)

## What are you seeing

```
  Error: /Stage[main]/Zabbix::Agent/Service[zabbix-agent]/ensure: change from 'stopped' to 'running' failed: Systemd start for zabbix-agent failed!
  journalctl log for zabbix-agent:
  -- Logs begin at Tue 2020-01-07 13:06:46 UTC, end at Tue 2020-01-07 13:10:16 UTC. --
  Jan 07 13:08:46 centos7-64-1 systemd[1]: Starting Zabbix Agent...
  Jan 07 13:08:46 centos7-64-1 zabbix_agentd[2119]: Starting Zabbix Agent [centos7-64-1]. Zabbix 2.4.8 (revision 59539).
  Jan 07 13:08:46 centos7-64-1 systemd[1]: New main PID 2119 does not belong to service, and PID file is not owned by root. Refusing.
  Jan 07 13:08:46 centos7-64-1 systemd[1]: New main PID 2119 does not belong to service, and PID file is not owned by root. Refusing.
  Jan 07 13:08:46 centos7-64-1 zabbix_agentd[2119]: using configuration file: /etc/zabbix/zabbix_agentd.conf
  Jan 07 13:08:46 centos7-64-1 zabbix_agentd[2119]: agent #0 started [main process]
  Jan 07 13:08:46 centos7-64-1 zabbix_agentd[2123]: agent #4 started [listener #3]
  Jan 07 13:08:46 centos7-64-1 zabbix_agentd[2121]: agent #2 started [listener #1]
  Jan 07 13:08:46 centos7-64-1 zabbix_agentd[2122]: agent #3 started [listener #2]
  Jan 07 13:08:46 centos7-64-1 zabbix_agentd[2120]: agent #1 started [collector]
  Jan 07 13:10:16 centos7-64-1 systemd[1]: zabbix-agent.service start operation timed out. Terminating.
[...]

PR #646, opened 2020-01-07
Pin CentOS acceptance tests 7.6.1810

baurmatt:
This is needed because of https://github.com/docker/for-linux/issues/835
and https://github.com/moby/moby/issues/38749.

Long story short: systemd on CentOS 7.7 is broken with current versions
of Docker.

#645


bastelfreak:
thanks for the fix!

PR #648, opened 2020-01-10
zabbix.com repository for Raspbian Linux

emetriqChris:
Zabbix.com provides a repository for Raspbian Linux under http://repo.zabbix.com/zabbix/<version>/raspbian/

The puppet-zabbix module uses the fact os.name for determining the URL of the used repository which is "Debian" for Raspbian Linux. The Debian repositiry is not providing "armhf" deb packages so the installation of zabbix fails.

This patch is an if condition which sets the variable $operatingsystem to 'raspbian' if the fact os.distro.id is 'Raspbian'. Otherwise the behaviour is not modified.

This way puppet-zabbix is choosing the right repository for installing zabbix on a Raspbian Linux.

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -123,7 +123,11 @@ class zabbix::repo (
             ,
           }
         } else {
-          $operatingsystem = downcase($facts['os']['name'])
+          if ($facts['os']['distro'][id] == 'Raspbian') {
+            $operatingsystem = 'raspbian'
+          } else {
+            $operatingsystem = downcase($facts['os']['name'])
+          }
           case $facts['os']['release']['full'] {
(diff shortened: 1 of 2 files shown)


bastelfreak:
thanks for the PR @emetriqChris. Can you please add Raspbian to the metadata.json?

emetriqChris:
Hi,

thanks for looking into this. I added Raspbian to metadata.json but somehow Travis crashed. It looks like the test couldn't login to the docker container. Maybe re-running could help.

Greetings,
Chris

bastelfreak:
I restarted the failed job

emetriqChris:
Thanks!

Issue #394, opened 2017-06-02
proxy and server manifest don't have interface for logtype


stanleyz:
There is no logtype parameter in server.pp and proxy.pp whereas the templates for these two support logtype.

PR #650, opened 2020-01-15
Add logtype parameter for zabbix server (issue #394)

dpavlotzky:


manifests/server.pp
```diff
@@ -310,6 +310,7 @@
   $nodeid                           = $zabbix::params::server_nodeid,
   $listenport                       = $zabbix::params::server_listenport,
   $sourceip                         = $zabbix::params::server_sourceip,
+  String $logtype                   = $zabbix::params::server_logtype,
```
bastelfreak: do you know if only specific values are allowed here? In that case you could use the Enum. Otherwise prohibit empty strings with `String[1]`.

bastelfreak:
Hi @dpavlotzky, thanks for the PR. Can you please take a look at the used email address? It isn't associated with your github account.

Is the log setup only specific for the server, or also for the agent? If so, the agent code would need to be updated as well.

dpavlotzky:
@bastelfreak I have updated the email address of my Github account. The agent already has the logtype parameter:
https://github.com/voxpupuli/puppet-zabbix/blob/fdd5fd13ce6bacf2e284d7768734de140492a7ab/manifests/params.pp#L272-L281
https://github.com/voxpupuli/puppet-zabbix/blob/fdd5fd13ce6bacf2e284d7768734de140492a7ab/manifests/agent.pp#L235

dpavlotzky:
The test build fails because a trailing comma is missing:
```
$ bundle exec rake $CHECK
manifests/server.pp:316:trailing_comma:WARNING:missing trailing comma after last element
The command "bundle exec rake $CHECK" exited with 1.
```
The complaint is about the following code:

https://github.com/dpavlotzky/puppet-zabbix/blob/afb8643095fca155d8e4df43af2ae9e935e60b35/manifests/server.pp#L313-L317

But the complaint is incorrect: adding a trailing comma here is a syntax error.

I will inline the Enum, but that is less pretty IMHO. Maybe someone can have a look at the test code to make it more robust.

PR #663, opened 2020-03-02
Fix type error on zabbix_proxy delete

baurmatt:
zbx.proxies.delete expects an array, we handed over a string. This fixes
it a adds proper tests.


baurmatt:
Travis for Ubuntu 16.04:

```
Error: Unable to connect to PostgreSQL server! (:5432)

  Error: /Stage[main]/Postgresql::Server::Service/Postgresql_conn_validator[validate_service_is_running]/ensure: change from 'absent' to 'present' failed: Unable to connect to PostgreSQL server! (:5432)
```

I *think* the error is unrelated because:
1. Works on my machine ;)
2. Happens on master as well
3. Works for all other OSs

PR #669, opened 2020-03-27
Fix zabbix::proxy idempotency

alexjfisher:
Active proxies don't use `ipaddress`, `use_ip` or `port`.  Specifically,
the zabbix_proxy provider will return `nil` for these properties for
active proxies.

### DIFF
--- a/manifests/proxy.pp
+++ b/manifests/proxy.pp
@@ -449,12 +449,20 @@ class zabbix::proxy (
   # is set to false, you'll get warnings like this:
   # "Warning: You cannot collect without storeconfigs being set"
   if $manage_resources {
-    class { 'zabbix::resources::proxy':
-      hostname  => $hostname,
-      ipaddress => $listen_ip,
-      use_ip    => $use_ip,
-      mode      => $mode,
-      port      => $listenport,
+    if String($mode) == '0' {
+      # Active proxies don't use `ipaddress`, `use_ip` or `port`.
+      class { 'zabbix::resources::proxy':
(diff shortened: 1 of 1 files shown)


alexjfisher:
This PR hasn't been tested by me.  I've created it following an issue reported by @tenajsystems on slack.

alexjfisher:
From slack...
> We have been using this zabbix module: https://forge.puppet.com/puppet/zabbix. Ever since we updated the module from version 6.7.0 to 8.0.0, each time puppet runs on the zabbix server, we get the below corrective changes. I have dug throw everything I can think of but no luck. I think this might be a bug with the module but I could be totally wrong so wanted to share it here just in case someone can point me in the right direction. In the Zabbix GUI ( Administration -> Proxies), there is no field to enter a port for the proxy when the proxy mode is Active. The port field is only available when the proxy mode is Passive so I believe the module is treating everything as Passive when it comes to the GUI. Any suggestions or thoughts on how to stop these corrective changes?
Notice: /Stage[main]/Zabbix::Resources::Web/Zabbix_proxy[zone1-zabbix-proxy]/port: defined 'port' as '10051' (corrective)
Notice: /Stage[main]/Zabbix::Resources::Web/Zabbix_proxy[zone2-zabbix-proxy]/port: defined 'port' as '10051' (corrective)

tenajsystems:
@alexjfisher I just tested this pull request and  your change worked like a charm! No more corrective changes. Thank you for fixing this bug! When will this be merged and the forge module updated?

PR #673, opened 2020-04-12
Use older xenial docker image

alexjfisher:
Needed because of https://github.com/docker/for-linux/issues/835

Also see 8496ed38e58e65385a1ee46c7240c49985474014


alexjfisher:
> Why do we need to use an older image?

To fix the tests.  Now that they're passing, I've updated the commit message and PR comment with explanation.

PR #664, opened 2020-03-02
Add acceptance testing for 4.2 and 4.4

baurmatt:
Test current version of Zabbix.

PR #674, opened 2020-04-13
Add Debian 10 support

bastelfreak:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,14 @@ matrix:
     bundler_args: --without development release
     env: PUPPET_INSTALL_TYPE=agent BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_debug=true BEAKER_setfile=debian9-64 BEAKER_HYPERVISOR=docker CHECK=beaker
     services: docker
+  - rvm: 2.5.3
+    bundler_args: --without development release
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_debug=true BEAKER_setfile=debian10-64 BEAKER_HYPERVISOR=docker CHECK=beaker
+    services: docker
+  - rvm: 2.5.3
+    bundler_args: --without development release
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_debug=true BEAKER_setfile=debian10-64 BEAKER_HYPERVISOR=docker CHECK=beaker
+    services: docker
 branches:
   only:
   - master
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ group :test do
   gem 'fakefs', '0.13.3',            :require => false if RUBY_VERSION < '2.3.0'
   gem 'fakefs',                      :require => false if RUBY_VERSION >= '2.3.0'
   gem 'zabbixapi',                   :require => false
+  gem 'rspec-puppet-facts',          :require => false, :git => 'https://github.com/ekohl/rspec-puppet-facts', :ref => '7674437d8d48af3a910ac3858c9eb63853dc1821'
 end
 
 group :development do
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -124,10 +124,12 @@ class zabbix::params {
     }
   }
 
-  # Zabbix overall params. Is used by all components.
-  $zabbix_version = downcase($facts['kernel']) ? {
-    'windows' => '4.4.5',
-    default   => '3.4',
+  if downcase($facts['kernel']) == 'windows' {
+    $zabbix_version = '4.4.5'
+  } elsif $facts['os']['name'] == 'Debian' and Integer($facts['os']['release']['major']) == 10 {
+    $zabbix_version =  '4.0'
+  } else {
+    $zabbix_version = '3.4'
   }
 
   $manage_startup_script = downcase($facts['kernel']) ? {
--- a/spec/acceptance/agent_spec.rb
+++ b/spec/acceptance/agent_spec.rb
@@ -1,9 +1,8 @@
 require 'spec_helper_acceptance'
 
 def agent_supported(version)
-  if default[:platform] =~ %r{(ubuntu-16.04|debian-9)-amd64}
-    return version != '2.4'
-  end
+  return version != '2.4' if default[:platform] =~ %r{(ubuntu-16.04|debian-9)-amd64}
+  return version >= '4.0' if default[:platform] =~ %r{debian-10-amd64}
   true
 end
 
@@ -14,7 +13,7 @@ end
       class { 'zabbix::agent':
         server               => '192.168.20.11',
         zabbix_package_state => 'latest',
-        zabbix_version       => '#{version}'
+        zabbix_version       => '#{version}',
       }
       EOS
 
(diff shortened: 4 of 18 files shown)


bastelfreak:
since the acceptance tests otherwise take ages, I tested the fix from @ekohl here.

.travis.yml
```diff
@@ -46,6 +46,14 @@ matrix:
     bundler_args: --without development release
     env: PUPPET_INSTALL_TYPE=agent BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_debug=true BEAKER_setfile=debian9-64 BEAKER_HYPERVISOR=docker CHECK=beaker
     services: docker
+  - rvm: 2.5.3
```
dhoppe, line 49: Where is the updated version of ```.sync.yml```?

spec/acceptance/agent_spec.rb
```diff
@@ -4,15 +4,22 @@
   if version == '2.4' && default[:platform] =~ %r{(ubuntu-16.04-amd64|debian-9-amd64)}
     next
   end
+  # Zabbix supports only Debian 10 on Zabbix 4
```
alexjfisher, line 7: This comment seems worded the wrong way round.  Should be 'Debian 10 only supports zabbix 4'?  What about 4.2 and 4.4?

spec/acceptance/agent_spec.rb
```diff
@@ -4,15 +4,22 @@
...
+  if default[:platform] =~ %r{debian-10-amd64}
+    next unless version == '4.0'
```
alexjfisher, line 9: Does this need tweaking now that https://github.com/voxpupuli/puppet-zabbix/pull/664 has been merged?

ekohl:
https://github.com/voxpupuli/puppet-zabbix/pull/676 may help to make it easier to manage.

spec/acceptance/agent_spec.rb
```diff
@@ -36,42 +36,53 @@ class { 'zabbix::agent':
...
-    describe file('/etc/zabbix/zabbix_agentd.conf') do
-      its(:content) { is_expected.to match %r{ListenIP=127.0.0.1} }
+    context 'With ListenIP set to an IP-Address' do
+      it 'works idempotently with no errors' do
+        pp = <<-EOS
+          class { 'zabbix::agent':
+            server               => '192.168.20.11',
+            zabbix_package_state => 'latest',
+            listenip             => '127.0.0.1',
+            zabbix_version       => '3.4'
```
ekohl, line 47: If you put this in the context and test it with every version then this should use the variable

spec/acceptance/agent_spec.rb
```diff
@@ -36,42 +36,53 @@ class { 'zabbix::agent':
...
-    describe file('/etc/zabbix/zabbix_agentd.conf') do
-      its(:content) { is_expected.to match %r{ListenIP=127.0.0.1} }
+    context 'With ListenIP set to lo' do
+      it 'works idempotently with no errors' do
+        pp = <<-EOS
+          class { 'zabbix::agent':
+            server               => '192.168.20.11',
+            zabbix_package_state => 'latest',
+            listenip             => 'lo',
+            zabbix_version       => '3.4'
```
ekohl, line 64: Same here

spec/acceptance/agent_spec.rb
```diff
@@ -36,42 +36,53 @@ class { 'zabbix::agent':
...
+          }
+          EOS
+        apply_manifest(pp, catch_failures: true)
+        apply_manifest(pp, catch_changes: true)
+      end
+      context 'With ListenIP set to an IP-Address' do
```
ekohl, line 70: Should this test also set the version?

manifests/params.pp
```diff
@@ -124,10 +124,14 @@
     }
   }
 
-  # Zabbix overall params. Is used by all components.
-  $zabbix_version = downcase($facts['kernel']) ? {
-    'windows' => '4.4.5',
-    default   => '3.4',
+
+
```
ekohl, line 128:
```suggestion
```

Gemfile
```diff
@@ -68,4 +71,6 @@ end
 ENV['PUPPET_VERSION'].nil? ? puppetversion = '~> 6.0' : puppetversion = ENV['PUPPET_VERSION'].to_s
 gem 'puppet', puppetversion, :require => false, :groups => [:test]
 
+gem 'irb'
```
ekohl, line 74: Debug

Gemfile
```diff
@@ -68,4 +71,6 @@ end
...
+gem 'beaker-vagrant'
```
ekohl, line 75: Should be in voxpupuli-acceptance

Gemfile
```diff
@@ -19,6 +19,9 @@ group :test do
   gem 'zabbixapi',                   :require => false
 end
 
+gem 'rspec-puppet-facts', :git => 'https://github.com/ekohl/rspec-puppet-facts', :branch => 'find-facter-version'
+gem 'facterdb', :git => 'https://github.com/camptocamp/facterdb', :branch => 'master'
```
ekohl, line 23: I'd prefer not to rely on git branches unless really needed. Perhaps link to the commit/pr that we need to be released?

spec/acceptance/agent_spec.rb
```diff
@@ -11,10 +13,11 @@ def agent_supported(version)
   describe "zabbix::agent class with zabbix_version #{version}", if: agent_supported(version) do
     it 'works idempotently with no errors' do
       pp = <<-EOS
+      include apt
```
ekohl, line 16: Why is including apt needed? It shouldn't be. There's this:
https://github.com/voxpupuli/puppet-zabbix/blob/87e09e3c31c28f0d6c58590f727b92ad5d1c8326/manifests/repo.pp#L98-L104
However, there's also https://github.com/puppetlabs/puppetlabs-apt/blob/04b6a898f8e6f733e1b4f62d3c1a1c631b10253c/manifests/source.pp#L71

I don't see why the repo class shouldn't unconditionally include it. It's not a literal class statement so it shouldn't be an issue. However, it defaults to true anyway so this should be redundant.

```suggestion
```

(repeated comment)

spec/acceptance/zabbix_application_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper_acceptance'
 require 'serverspec_type_zabbixapi'
 
-describe 'zabbix_application type' do
+describe 'zabbix_application type', unless: default[:platform] =~ %r{debian-10-amd64} do
```
ekohl, line 4: Rather than skipping, should we just not set the `zabbix_version` and use the default in this test suite (repeated in multiple files)?

Gemfile
ekohl: 1.3.0 was released so this can be removed.

spec/acceptance/zabbix_application_spec.rb
bastelfreak: sadly that won't work. the custom types use the zabbixapi rubygem and that currently only supports older zabbix versions

ekohl: https://github.com/voxpupuli/puppet-zabbix/pull/623 suggests this can work if we update it.

spec/acceptance/zabbix_application_spec.rb
```diff
@@ -38,7 +38,15 @@ class { 'zabbix':
     # setup zabbix. Apache module isn't idempotent and requires a second run
     it 'works with no error on the first apply' do
       # Cleanup old database
-      shell('/opt/puppetlabs/bin/puppet resource service zabbix-server ensure=stopped; /opt/puppetlabs/bin/puppet resource package zabbix-server-pgsql ensure=purged; rm -f /etc/zabbix/.*done; su - postgres -c "psql -c \'drop database if exists zabbix_server;\'"')
+      cleanup_script = <<-SHELL
```
ekohl, line 41: Can we extract this somewhere to avoid the repetition?

bastelfreak:
@baurmatt any idea how we can get the tests working on CentOS 7 / any idea why they don't fail on master?
During the tests, server_spec.rb installs zabbix 4.4, which works. afterwars zabbix_application_spec.rb gets loaded which installs 3.4: https://travis-ci.org/github/voxpupuli/puppet-zabbix/jobs/677866025#L8857

Somehow the cache is messed up and yum wants to install the 4.4 version of the zabbix server:
https://travis-ci.org/github/voxpupuli/puppet-zabbix/jobs/677866025#L8887

* I'm not sure why this is happening now
* It even happens with the `shell('yum clean all') if fact('os.family') == 'RedHat'` snippet that's also executed:
```
centos7-64-1 20:23:33$ yum clean all
  Loaded plugins: fastestmirror, ovl
  Cleaning repos: Zabbix_7_x86_64 Zabbix_nonsupported_7_x86_64 base extras puppet5
                : updates
  Cleaning up list of fastest mirrors
centos7-64-1 executed in 0.21 seconds
```
* Why does master pass??

spec/acceptance/zabbix_application_spec.rb
bastelfreak: I would like to get this merged and work on #623 afterwards

bastelfreak:
after adding way more yum cleanup statements, it turned green \o/

PR #677, opened 2020-04-16
adding possibility to define db_schema in zabbix.conf.php template

sasskinn12:


bastelfreak:
Hi @sasskinn12, thanks for the PR. Can you please take a look at the used email address in your commit? It isn't associated with your github account.

Please also add a test that ensures setting the variable puts it correctly into the file.

anthonysomerset:
this is actually clone/fixed by #695

PR #678, opened 2020-04-16
Allow mode changing for zabbix_proxy

baurmatt:
It is currently not possible to change the mode of an existing Zabbix proxy. This changes it.

### DIFF
--- a/lib/puppet/provider/zabbix_proxy/ruby.rb
+++ b/lib/puppet/provider/zabbix_proxy/ruby.rb
@@ -16,6 +16,7 @@ Puppet::Type.type(:zabbix_proxy).provide(:ruby, parent: Puppet::Provider::Zabbix
       # p['interface'] is a Hash if the host is a passive proxy
       new(
         ensure: :present,
+        id: p['proxyid'].to_i,
         name: p['host'],
         ipaddress: p['interface'].is_a?(Hash) ? p['interface']['ip'] : nil,
         use_ip: p['interface'].is_a?(Hash) ? p['interface']['use_ip'] : nil,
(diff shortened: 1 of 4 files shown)


lib/puppet/provider/zabbix_proxy/ruby.rb
```diff
@@ -60,12 +61,46 @@ def create
...
+  # it might not be there if we switch the proxy from active to passive.
+  def update_mode_interface
+    zbx.proxies.update(
+      proxyid: @property_hash[:id],
+      status: @resource[:mode] + 5, # Normally 0 is active and 1 is passive, in the API, its 5 and 6
+      interface: {
+        ip: @resource[:ipaddress],
+        dns: @resource[:hostname],
+        useip: @resource[:use_ip] ? 1 : 0,
+        port: @resource[:port]
```
alexjfisher, line 86: What happens if eg `@resource[:port]` is `undef`?  Does it need to be `port: @resource[:port] || @property_hash[:port]`

lib/puppet/provider/zabbix_proxy/ruby.rb
```diff
@@ -60,12 +61,46 @@ def create
...
   end
 
+  #
+  # zabbix_proxy properties
+  #
   mk_resource_methods
+
+  # We have to update all parameter of the interface and mode because
+  # it might not be there if we switch the proxy from active to passive.
+  def update_mode_interface
```
alexjfisher, line 78: If you made this method `def flush` instead, would you even need the property setters?

baurmatt:
Closed in favor of #679.

PR #679, opened 2020-04-17
Improve `zabbix_proxy` type

alexjfisher:
- It is now possible to manage individual `zabbix_proxy` resource properties.
- The type has been improved to do enhanced validation of all properties.
- Switching between active and passive proxies (and vice-versa) is now supported.
- Improved acceptance tests (Thanks @baurmatt )
- Improved type unit tests.

### DIFF
--- a/lib/puppet/provider/zabbix_proxy/ruby.rb
+++ b/lib/puppet/provider/zabbix_proxy/ruby.rb
@@ -2,26 +2,16 @@ require_relative '../zabbix'
 Puppet::Type.type(:zabbix_proxy).provide(:ruby, parent: Puppet::Provider::Zabbix) do
   confine feature: :zabbixapi
 
-  def self.instances
-    proxies = zbx.query(
-      method: 'proxy.get',
-      params: {
-        output: 'extend',
-        selectInterface: %w[interfaceid type main ip port useip]
-      }
-    )
+  def initialize(value = {})
+    super(value)
+    @property_flush = {}
+  end
 
+  mk_resource_methods
+
+  def self.instances
     proxies.map do |p|
-      # p['interface'] is an Array if the host is a active proxy
-      # p['interface'] is a Hash if the host is a passive proxy
-      new(
-        ensure: :present,
-        name: p['host'],
-        ipaddress: p['interface'].is_a?(Hash) ? p['interface']['ip'] : nil,
-        use_ip: p['interface'].is_a?(Hash) ? p['interface']['use_ip'] : nil,
-        mode: p['status'].to_i - 5,
-        port: p['interface'].is_a?(Hash) ? p['interface']['port'] : nil
-      )
+      new(proxy_properties_hash(p))
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/puppet/provider/zabbix_proxy/ruby.rb
```diff
@@ -33,39 +23,178 @@ def self.prefetch(resources)
...
+      }
+    else
+      interface = nil
+    end
+
+    zbx.proxies.create(
+      {
+        host: @resource[:hostname],
+        status: self.class.mode_to_status(mode),
+        interface: interface
```
baurmatt, line 76: haha :D I was about to open an issue for this. `interfaces` is deprecated as of Zabbix 2.2 in favor of `interface` and was removed in 3.4. Thank you! 👍

lib/puppet/provider/zabbix_proxy/ruby.rb
```diff
@@ -33,39 +23,178 @@ def self.prefetch(resources)
...
+  def mode=(value)
+    @property_flush[:mode] = value
+  end
+
+  def change_to_passive_proxy
+    Puppet.debug("We're changing to a passive proxy")
+    # We need to call zbx.proxies.update with an `interface` hash.
+
+    useip = if @resource[:use_ip].nil?
+              1 # Default to true
```
baurmatt, line 113: Should we just default this on the at the type?

lib/puppet/provider/zabbix_proxy/ruby.rb
```diff
@@ -33,39 +23,178 @@ def self.prefetch(resources)
...
+            else
+              @resource[:use_ip] == :true ? 1 : 0
+            end
+
+    interface = {
+      ip: @resource[:ipaddress] || @property_hash[:ipaddress] || '127.0.0.1',
```
baurmatt, line 119: Should we use `127.0.0.1` as a default at the type?

lib/puppet/provider/zabbix_proxy/ruby.rb
alexjfisher: No, otherwise the property is always managed.

lib/puppet/provider/zabbix_proxy/ruby.rb
alexjfisher: My starting point was your code, so you can take credit for fixing this! :)

lib/puppet/provider/zabbix_proxy/ruby.rb
alexjfisher: > The defaultto method is used to specify default values for any attribute. Defaults are typically only specified for parameters. They are less common for properties because a default value will result in that property always being managed (there is no way to unset a value for the property).

From https://www.amazon.com/Puppet-Types-Providers-Extending-Ruby/dp/1449339328

baurmatt: Interesting 🤔

PR #680, opened 2020-04-18
modulesync 2.12.0 / zabbix::web: Allow httpd to speak to the database

bastelfreak:
modulesync 2.12.0

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,5 +1,11 @@
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
-configure_beaker
+configure_beaker do |host|
+  # The CentOS docker image has a yum config that won't install docs, to keep used space low
+  # zabbix packages their SQL file as doc, we need that for bootstrapping the database
+  if host[:platform] =~ %r{el-7-x86_64} && host[:hypervisor] =~ %r{docker}
+    on(host, "sed -i '/nodocs/d' /etc/yum.conf")
+  end
+end
 
 Dir['./spec/support/acceptance/**/*.rb'].sort.each { |f| require f }
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,3 +1,5 @@
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
 configure_beaker
+
+Dir['./spec/support/acceptance/**/*.rb'].sort.each { |f| require f }
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,21 +1,3 @@
-require 'beaker-rspec'
-require 'beaker-puppet'
-require 'beaker/puppet_install_helper'
-require 'beaker/module_install_helper'
+require 'voxpupuli/acceptance/spec_helper_acceptance'
 
-run_puppet_install_helper unless ENV['BEAKER_provision'] == 'no'
-install_ca_certs unless ENV['PUPPET_INSTALL_TYPE'] =~ %r{pe}i
-install_module_on(hosts)
-install_module_dependencies_on(hosts)
-
-RSpec.configure do |c|
-  # Readable test descriptions
-  c.formatter = :documentation
-  hosts.each do |host|
-    if host[:platform] =~ %r{el-7-x86_64} && host[:hypervisor] =~ %r{docker}
-      on(host, "sed -i '/nodocs/d' /etc/yum.conf")
-    end
-  end
-end
-
-Dir['./spec/support/acceptance/**/*.rb'].sort.each { |f| require f }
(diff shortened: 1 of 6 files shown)


bastelfreak:
so:
* spec/acceptance/server_spec.rb passes locally for me
* I can reduce the 4 errors in spec/acceptance/zabbix_application_spec.rb to 3 with this spatch:

```patch
diff --git a/spec/acceptance/zabbix_application_spec.rb b/spec/acceptance/zabbix_application_spec.rb
index 61129f8..b6d6d65 100644
--- a/spec/acceptance/zabbix_application_spec.rb
+++ b/spec/acceptance/zabbix_application_spec.rb
@@ -38,7 +38,15 @@ describe 'zabbix_application type' do
     # setup zabbix. Apache module isn't idempotent and requires a second run
     it 'works with no error on the first apply' do
       # Cleanup old database
-      shell('/opt/puppetlabs/bin/puppet resource service zabbix-server ensure=stopped; /opt/puppetlabs/bin/puppet resource package zabbix-server-pgsql ensure=purged; rm -f /etc/zabbix/.*done; su - postgres -c "psql -c \'drop database if exists zabbix_server;\'"')
+      cleanup_script = <<-SHELL
+      /opt/puppetlabs/bin/puppet resource service zabbix-server ensure=stopped
+      /opt/puppetlabs/bin/puppet resource package zabbix-server-pgsql ensure=purged
+      rm -f /etc/zabbix/.*done
+      if id postgres > /dev/null 2>&1; then
+        su - postgres -c "psql -c 'drop database if exists zabbix_server;'"
+      fi
+      SHELL
+      shell(cleanup_script)
 
       apply_manifest(pp1, catch_failures: true)
     end

```

(which I would like to refactor into a method so all acceptance test files can use it)

bastelfreak:
I restarted the last successful merge to master build: https://travis-ci.org/github/voxpupuli/puppet-zabbix/jobs/675009991

bastelfreak:
whyyyyyyyyyyyyyyyyyyyyyyyyyyy does it pass

bastelfreak:
I created a simple PR against master that only touches our README.md: https://github.com/voxpupuli/puppet-zabbix/pull/681

bastelfreak:
which passed. Now I cherry-picked the cleanup commit into another PR:
https://github.com/voxpupuli/puppet-zabbix/pull/682

bastelfreak:
@baurmatt I'm still debugging this, but if you've any ideas why this works on master but fails here, please let me know :)

bastelfreak:
I've no clue why the code in docker fails to create a database for one zabbix version:
https://travis-ci.org/github/voxpupuli/puppet-zabbix/jobs/679220993#L9204


I'm unable to reproduce this locally with vagrant

baurmatt:
@bastelfreak Ha! I've seen something similar with Ubuntu 18.04. Turned out that it ships a `/etc/dpkg/dpkg.cfg.d/excludes` which includes the following directories by default:

```
# Drop all man pages
path-exclude=/usr/share/man/*

# Drop all documentation ...
path-exclude=/usr/share/doc/*

# ... except copyright files ...
path-include=/usr/share/doc/*/copyright

# ... and Debian changelogs
path-include=/usr/share/doc/*/changelog.Debian.*
```

This led to /usr/share/doc/zabbix-*-pgsql-3.4*/*.sql not being installed and hence failing tests.

I think the problem here is, that `nodocs` isn't removed anymore from /etc/yum.conf. That's probably the equivalent in CentOS.

spec/spec_helper_acceptance.rb
```diff
@@ -1,21 +1,5 @@
...
-install_ca_certs unless ENV['PUPPET_INSTALL_TYPE'] =~ %r{pe}i
-install_module_on(hosts)
-install_module_dependencies_on(hosts)
-
-RSpec.configure do |c|
-  # Readable test descriptions
-  c.formatter = :documentation
-  hosts.each do |host|
-    if host[:platform] =~ %r{el-7-x86_64} && host[:hypervisor] =~ %r{docker}
-      on(host, "sed -i '/nodocs/d' /etc/yum.conf")
```
baurmatt, line 16: @bastelfreak This shitty thing! :D

ekohl: So we need this? What breaks?

baurmatt: I leads to files in /usr/share/doc/* not being installed. As Zabbix places their database initialization script there, this breaks our acceptance CI.

ekohl: Ah yes, I now see the IRC conversation. Pointing @bastelfreak in the right direction.

bastelfreak: Thanks @baurmatt . We figured the same out in parallel yesterday on IRC. The current code should now work again on docker images. Are you already working on Ubuntu 18/20 support for the module?

PR #681, opened 2020-04-19
please mdlinter in README.md

bastelfreak:
this fixes:
`MD001 Header levels should only increment by one level at a time`


bastelfreak:
it passes :thinking:

baurmatt:
Should we modulesync a matching Travis job for this? We use the following internally in our Gitlab-CI:

```yaml
markdown-lint:
  stage: test
  image: node:alpine
  before_script:
    - npm install -g markdownlint markdownlint-cli
  script: find . -type f -name "*\.md" -print0 | xargs -0 markdownlint
```

bastelfreak:
in general I would like to have this. I'm using https://rubygems.org/gems/mdl at the moment. The issue is that none of our README.md passes, so enforcing this in a CI pipeline is tricky :(

baurmatt:
@bastelfreak Can we just trigger a warning instead of a error? This might gradually fix it.

PR #682, opened 2020-04-19
Cleanup and fix acceptance test helper

bastelfreak:

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -17,3 +17,5 @@ RSpec.configure do |c|
     end
   end
 end
+
+Dir['./spec/support/acceptance/**/*.rb'].sort.each { |f| require f }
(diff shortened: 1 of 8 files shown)


spec/support/acceptance/cleanup_zabbix.rb
```diff
@@ -0,0 +1,11 @@
+def cleanup_zabbix
+  cleanup_script = <<-SHELL
+  /opt/puppetlabs/bin/puppet resource service zabbix-server ensure=stopped
+  /opt/puppetlabs/bin/puppet resource package zabbix-server-pgsql ensure=purged
+  rm -f /etc/zabbix/.*done
+  if id postgres > /dev/null 2>&1; then
```
bastelfreak, line 6: I labeled this as bug because in the past, the code didn't check if the user exists. The spec tests worked by accident because the first test added the user and the other depended on them. Tests in the past failed if you tried to execute them standalone.

spec/spec_helper_acceptance.rb
```diff
@@ -17,3 +17,5 @@
     end
   end
 end
+
+Dir['./spec/support/acceptance/**/*.rb'].sort.each { |f| require f }
```
bastelfreak, line 21: I will add that to modulesync_config

PR #688, opened 2020-05-03
Drop ruby < 2.3 fakefs gem dependency

ekohl:
This is no longer needed since we don't test on those old Ruby versions anymore.


alexjfisher:
Build failed due to expired certificates on the apt and yum repos.  Looks like these were updated this morning and hopefully the restarted build will now pass.

ekohl:
Thanks! Green now.

PR #691, opened 2020-05-27
opsgenie integration plugin needs ProtectHome = no

veninga:

For opsgenie integration plugins it is needed that on systemd enabled systems the setting "ProtectHome" is "no". The reason is that the integration plugin is tightly coupled in /home/opsgenie.


bastelfreak:
Hi,
I think yes is a very sane default value here. can you please make this option configureable and keep the default to yes?

PR #694, opened 2020-06-09
Zabbix database creation on zabbix version 5.x

StanHVA:
Zabbix 5.x is out. This means that the regex in the database creation files is no longer valid (it checks for v3/4). It needs to include version 5 as well.

Zabbix database creation on version 5.x


bastelfreak:
thanks for the PR! Can you update the acceptance tests in spec/acceptance/ to also test version 5?

PR #699, opened 2020-08-02
modulesync 3.0.0 & puppet-lint fixes & rubocop fixes

bastelfreak:

### DIFF
--- a/spec/acceptance/zabbix_application_spec.rb
+++ b/spec/acceptance/zabbix_application_spec.rb
@@ -1,6 +1,7 @@
 require 'spec_helper_acceptance'
 require 'serverspec_type_zabbixapi'
 
+# rubocop:disable RSpec/LetBeforeExamples
 describe 'zabbix_application type', unless: default[:platform] =~ %r{debian-10-amd64} do
   context 'create zabbix_application resources' do
     # This will deploy a running Zabbix setup (server, web, db) which we can
--- a/spec/acceptance/zabbix_application_spec.rb
+++ b/spec/acceptance/zabbix_application_spec.rb
@@ -38,7 +38,7 @@ describe 'zabbix_application type', unless: default[:platform] =~ %r{debian-10-a
     # setup zabbix. Apache module isn't idempotent and requires a second run
     it 'works with no error on the first apply' do
       # Cleanup old database
-      cleanup_zabbix
+      prepare_host
 
       apply_manifest(pp1, catch_failures: true)
     end
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,11 +1,6 @@
+# This file is completely managed via modulesync
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
-configure_beaker do |host|
-  # The CentOS docker image has a yum config that won't install docs, to keep used space low
-  # zabbix packages their SQL file as doc, we need that for bootstrapping the database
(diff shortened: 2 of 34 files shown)


dhollinger:
@bastelfreak Could you look into the failing tests?

bastelfreak:
oh why doesn't centos apply anymore :(

ekohl:
Was there a version change?

```
Notice: /Stage[main]/Zabbix::Database::Postgresql/Exec[zabbix_server_create.sql]/returns: /bin/sh: line 0: cd: /usr/share/doc/zabbix-*-pgsql-3.0*/: No such file or directory
```

spec/spec_helper_acceptance.rb
```diff
@@ -1,11 +1,6 @@
+# This file is completely managed via modulesync
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
-configure_beaker do |host|
-  # The CentOS docker image has a yum config that won't install docs, to keep used space low
```
baurmatt, line 4: Hey @bastelfreak, I think removing those lines might be the reason why the tests fails.

bastelfreak: oh right, I can even remember that people debugged that on IRC. the docs packages contain the sql files.

spec/spec_helper_acceptance.rb
```diff
@@ -1,11 +1,6 @@
+# This file is completely managed via modulesync
```
ekohl, line 1: I'd keep it as it was. The default is really only 2 lines so IMHO there's no benefit in syncing it. Much cleaner to just do the node setup once than to do it in every single test.

ekohl (changes requested):

spec/acceptance/zabbix_application_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'spec_helper_acceptance'
 require 'serverspec_type_zabbixapi'
 
+# rubocop:disable RSpec/LetBeforeExamples
```
alexjfisher, line 4: Did this need to be disabled?

bastelfreak: I assume so. Rubocop complained that this line is now useless and needs to be removed.

bastelfreak: ah, wrong comment. yes, rubocop said there are multiple issues for that cop. I wasn't able to fix this quickly and we really need to get this modulesync merged, so I disabled it.

dhollinger (approved):

spec/support/acceptance/prepare_host.rb
```diff
@@ -1,5 +1,10 @@
-def cleanup_zabbix
-  shell('rm -rf /etc/yum.repos.d/Zabbix*.repo; rm -rf /var/cache/yum/x86_64/*/Zabbix*; yum clean all --verbose') if fact('os.family') == 'RedHat'
+def prepare_host
+  if fact('os.family') == 'RedHat'
+    shell('rm -rf /etc/yum.repos.d/Zabbix*.repo; rm -rf /var/cache/yum/x86_64/*/Zabbix*; yum clean all --verbose')
+    if fact('os.release.major').to_i == 7
+      shell('sed -i "/nodocs/d" /etc/yum.conf')
+    end
```
ekohl, line 6: I really wanted the comment to survive.

bastelfreak: oh right, my bad. fixed it in https://github.com/voxpupuli/puppet-zabbix/pull/710

PR #700, opened 2020-08-05
Purging of Templates flag, IPMI Interface support, PSK Authentication

cbergmann:
This Pull request implements three Features:

1. It adds a new parameter `purge_templates` to zabbix_agent that controls if additional templates configured in the server for the zabbix_host resource should be deleted.
2. it implements psk authentication for zabbix_host resources
3. it fixes a problem with zabbix_host resources with only IPMI interfaces 

I found no easy way to split these commits into three PRs. If you know how to do this I would highly appreciate this.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -356,11 +356,16 @@ class zabbix::agent (
   # to network name. If more than 1 interfaces are available, we
   # can find the ipaddress of this specific interface if listenip
   # is set to for example "eth1" or "bond0.73".
-  $listen_ip = $listenip ? {
-    /^(e|lo|bond|lxc|tap|tun|virbr).*/ => fact("networking.interfaces.${listenip}.ip"),
-    '*' => undef,
-    default => $listenip,
+  $listen_array = flatten([$listenip])
+  $listen_ips = $listen_array.map |$listenip| {
+    $listenip ? {
+      /^(e|lo|bond|lxc|tap|tun|virbr).*/ => fact("networking.interfaces.${listenip}.ip"),
+      '*' => undef,
+      default => $listenip,
+    }
   }
+  $listen_ip = join($listen_ips,',')
+
 
   # If the user specified a psk but no file we will use the default filename.
   # If only the file is specified we configure it.
@@ -415,7 +420,7 @@ class zabbix::agent (
 
     class { 'zabbix::resources::agent':
       hostname         => $_hostname,
-      ipaddress        => $listen_ip,
+      ipaddress        => $listen_ips[0],
       use_ip           => $agent_use_ip,
       port             => $listenport,
       groups           => [$groups].flatten(),
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -323,6 +323,7 @@ class zabbix::agent (
   String $additional_service_params               = $zabbix::params::additional_service_params,
   String $service_type                            = $zabbix::params::service_type,
   Boolean $manage_startup_script                  = $zabbix::params::manage_startup_script,
+  Optional[Boolean] $purge_templates              = $zabbix::params::agent_purge_templates,
 ) inherits zabbix::params {
 
   # the following two codeblocks are a bit blargh. The correct default value for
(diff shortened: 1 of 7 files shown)


manifests/agent.pp
```diff
@@ -293,10 +305,13 @@
   $tlsaccept                                      = $zabbix::params::agent_tlsaccept,
   $tlscafile                                      = $zabbix::params::agent_tlscafile,
   $tlscertfile                                    = $zabbix::params::agent_tlscertfile,
+  $tlscertissuer                                  = $zabbix::params::agent_tlscertissuer,
```
bastelfreak, line 308: can you add datatypes for those new parameters please?

manifests/agent.pp
```diff
@@ -308,6 +323,7 @@
   String $additional_service_params               = $zabbix::params::additional_service_params,
   String $service_type                            = $zabbix::params::service_type,
   Boolean $manage_startup_script                  = $zabbix::params::manage_startup_script,
+  Optional[Boolean] $purge_templates              = $zabbix::params::agent_purge_templates,
```
bastelfreak, line 326: Optional around a boolean is an anti pattern that we try to avoid. Can't we just default to true or false?

bastelfreak: I think this parameter is also missing in the documentation?

manifests/agent.pp
```diff
@@ -340,10 +356,38 @@
...
-    default => $listenip,
+  $listen_array = flatten([$listenip])
+  $listen_ips = $listen_array.map |$listenip| {
+    $listenip ? {
+      /^(e|lo|bond|lxc|tap|tun|virbr).*/ => fact("networking.interfaces.${listenip}.ip"),
+      '*' => undef,
+      default => $listenip,
+    }
+  }
+  $listen_ip = join($listen_ips,',')
```
bastelfreak, line 367: I think those changed lines don't really relate to the rest of this PR? Can you provide it as a dedicated PR?

manifests/agent.pp
```diff
@@ -340,10 +356,38 @@
...
+
+
+  # If the user specified a psk but no file we will use the default filename.
+  # If only the file is specified we configure it.
+  # If neither file nor psk is specified nothing is configured.
+  if $tlspsk and ! $tlspskfile {
+    $real_tlspskfile = $zabbix::params::default_pskfile_path
```
bastelfreak, line 374: what's the benefit of having default_pskfile_path and tlspskfile? can't we just set zabbix::agent::tlspskfile to the value you assigned in params.pp to default_pskfile_path?

bastelfreak:
thanks for the awesome PR! I left some inline comments. I'm afraid we need to get https://github.com/voxpupuli/puppet-zabbix/pull/699 green somehow first. That will fix most of your CI failures. I'm very familiar with Puppet, but not  that much with Zabbix. If you've any clue with https://github.com/voxpupuli/puppet-zabbix/pull/699 fails please respond there. that would unblock many other PRs as well.

igalic (changes requested):
i don't understand why this isn't three separate pull requests

github's review UI makes it impossible to review a pull request commit by commit, so packing three such vastly different features and fixes into a single pull request makes review from people who aren't deeply familiar with the code and the server near impossible.

aclarkee (commented):
Testing this in my dev environment, (For PSK support & IPMI) so have added some thoughts/things I came across.

aclarkee:
FYI @cbergmann as of zabbix 5.4 - PSK values are write only. Current logic triggers a refresh of the PSK values on every run, since they can't be read back to verify. 

Any chance you've fixed that in your environment?

cbergmann:
> 
> 
> FYI @cbergmann as of zabbix 5.4 - PSK values are write only. Current logic on 5.4.x triggers a refresh of the PSK values on every run, since they can't be read back to verify.
> 
> Any chance you've fixed that in your environment?

Unfortunately I am currently not using this Module. I forwarded this information to my former colleagues and will come back to you if they had found a workaround.

aclarkee:
> > FYI @cbergmann as of zabbix 5.4 - PSK values are write only. Current logic on 5.4.x triggers a refresh of the PSK values on every run, since they can't be read back to verify.
> > Any chance you've fixed that in your environment?
> 
> Unfortunately I am currently not using this Module. I forwarded this information to my former colleagues and will come back to you if they had found a workaround.

Thanks - I have made a workaround myself, but it's pretty much a bodge - and not really worthy of being merged in..

Would like to see a better fix from someone who better understands ruby/puppet exported resources :)

PR #701, opened 2020-08-18
have agent service provider default to fact $::service_provider

automaticserver:
The provider for the agent service resource should default to the fact $::service_provider rather than undef. The fact is automatically provided by Facter. The default value can be overridden by creating a custom fact, e.g. when a different init system has been installed.

n/a


igalic:
the failing tests could probably be fixed with a module_sync

manifests/agent.pp
```diff
@@ -440,7 +440,7 @@
     $service_provider = 'init'
     $service_path = '/etc/rc.d/init.d'
   } else {
-    $service_provider = undef
+    $service_provider = $::service_provider
```
mmerfort, line 443: Please use the facts hash for this fact: `$facts['service_provider']`

bastelfreak:
Why should we explicitly set the provider? Puppet already discovers the provider. Did you run into any specific error that you would like to solve?

ansgarwiechers:
> Why should we explicitly set the provider? Puppet already discovers the provider. Did you run into any specific error that you would like to solve?
We're currently migrating servers to the s6 init system, and our custom provider is not automatically detected by Puppet. Normally this wouldn't be an issue b/c we can override the default provider for service resources, or (with most modules where this doesn't work) can at least specify the provider via a module parameter. However, the Zabbix modules does not have a parameter for overriding the provider, and ignores the default provider override b/c it explicitly passes `undef` to the service resource.

bastelfreak:
what about moving the variable service_provider from an internal variable to a parameter?

ansgarwiechers:
That would be another option, and we can do that if that's what you prefer. It requires more code and an interface change, though. Assigning `$facts['service_provider']` was the less intrusive change IMO.

ansgarwiechers:
So, can we go with the minimum-impact approach? Or should we change the PR to introduce a parameter `service_provider`? In case of the latter, should the default value remain `undef`? And should the parameter be restricted to the class `zabbix::agent` or should it also be added to the parent class too?

bastelfreak:
please introduce the new parameter. It should default to undef and also be added to the parent class.

igalic (changes requested):
can you please add spec tests for these new parameters?
if we have spec tests testing any of their surroundings, that is…

ansgarwiechers:
> can you please add spec tests for these new parameters?
> if we have spec tests testing any of their surroundings, that is…

Something like this?
```
context 'when declaring custom provider' do
  let :params do
    {
      service_provider: "#{facts['service_provider']}"
    }
  end

  it { is_expected.to contain_service(service_name).with(
    provider: "#{facts['service_provider']}"
  )}
end
```

igalic:
yes, that test case looks sensible

********

i have a general question:

what happens when, in your Profile or your site.pp, you set a (global) default for service provider to s6?

```puppet
Service {
  provider => 's6',
}
```

ansgarwiechers:
> what happens when, in your Profile or your site.pp, you set a (global) default for service provider to s6?
> 
> ```puppet
> Service {
>   provider => 's6',
> }
> ```
> 
> shouldn't the `undef` case then default to this arrow_up setting?

That is what we're doing. However, the Zabbix agent class explicitly passes a value to the provider parameter (even if that value is `undef`):

```
service { $servicename:
  ...
  provider   => $service_provider,
  ...
}
```

Because of that the global default value is ignored.

manifests/agent.pp
```diff
@@ -433,21 +438,26 @@
...
   # Controlling the 'zabbix-agent' service
-  if str2bool(getvar('::systemd')) {
-    $service_provider = 'systemd'
-    $service_path = undef
-  } elsif $facts['os']['name'] == 'AIX' {
-    $service_provider = 'init'
-    $service_path = '/etc/rc.d/init.d'
+  if $service_provider == undef {
+    if str2bool(getvar('::systemd')) {
+      $real_service_provider = 'systemd'
```
ekohl, line 443: This is a bit odd. Essentially this gets `$facts['systemd']` but that is defined as `service_provider == 'systemd'` (https://github.com/camptocamp/puppet-systemd/blob/03d94faacfe51ce95e8ab048f79cfd2617089b36/lib/facter/systemd.rb#L39).

So really, this whole systemd specific block can be removed and handled by the `else` case.

Given that, is the whole change still needed?

igalic: @bastelfreak and i have been discussing this on IRC, and it really is an odd case: this code shouldn't exist, and this patch shouldn't be necessary

ekohl: @automaticserver could you instead write a patch that removes this code and doesn't introduce a `service_provider` parameter? I don't know about the AIX code, but at least systemd should be removed here. It's probably cleaner as a separate PR.

ansgarwiechers: > could you instead write a patch that removes this code and doesn't introduce a service_provider parameter? I don't know about the AIX code, but at least systemd should be removed here. It's probably cleaner as a separate PR.

No.

If the AIX code remains then the module would still explicitly assign `undef` to the `provider` parameter of the service resourece, which would still supersede the default value we're assigning in the global scope. Meaning the problem we've been trying to resolve for three weeks now (and which could've easily been solved by a one-line change) would still persist.

ekohl: I quickly hacked up https://github.com/voxpupuli/puppet-zabbix/pull/711 which I think will solve it. Would be good to get some verification on that though.

igalic:
i'd like to close this now that #711 is merged

@ansgarwiechers, @automaticserver
can you please verify that works for you?

ansgarwiechers:
> i'd like to close this now that #711 is merged
> 
> can you please verify that works for you?

Yes, that works for us.

PR #665, opened 2020-03-06
Support for zabbix::agent on AIX

bjschafer:
#### Pull Request (PR) description

This initial pass offers support for installing and managing the Zabbix
agent on AIX systems. It has been tested against AIX 7.1 and 7.2.

No support is provided for managing a repo; package installation support
depends on a `yum` repo being available on the machine.

#### This Pull Request (PR) fixes the following issues
Fixes #504

PR #711, opened 2020-09-09
Reduce configuration on agent service

ekohl:
This avoids setting parameters as much as possible.

hasstatus and hasrestart have defaulted to true since Puppet 2.7 so there's no need to set those.

The systemd fact is defined as $fact['service_provider'] == 'systemd' so there's no need to set that either.

Then there's AIX left. It sets it this was because it allows users to override it in site.pp via service defaults. It now only sets the service_provider and not the path. This path is set in the provider since [Puppet 4.6.0](https://github.com/puppetlabs/puppet/commit/34e45ab3f562d1029bfeaafe9a11757ea3b3e92c).

I didn't verify this and only did it by theory so review carefully.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -432,24 +432,19 @@ class zabbix::agent (
   }
 
   # Controlling the 'zabbix-agent' service
-  if str2bool(getvar('::systemd')) {
-    $service_provider = 'systemd'
-    $service_path = undef
-  } elsif $facts['os']['name'] == 'AIX' {
-    $service_provider = 'init'
-    $service_path = '/etc/rc.d/init.d'
-  } else {
-    $service_provider = undef
-    $service_path = undef
-  }
   service { $servicename:
-    ensure     => $service_ensure,
-    enable     => $service_enable,
-    path       => $service_path,
-    provider   => $service_provider,
-    hasstatus  => true,
-    hasrestart => true,
-    require    => Package[$zabbix_package_agent],
+    ensure  => $service_ensure,
+    enable  => $service_enable,
+    require => Package[$zabbix_package_agent],
+  }
+
+  # Override the service provider on AIX
+  # Doing it this way allows overriding it on other platforms
+  if $facts['os']['name'] == 'AIX' {
+    Service[$servicename] {
+      service_provider => 'init',
+      service_path     => '/etc/rc.d/init.d',
+    }
   }
 
   # Configuring the zabbix-agent configuration file
--- a/spec/classes/agent_spec.rb
+++ b/spec/classes/agent_spec.rb
@@ -68,24 +68,21 @@ describe 'zabbix::agent' do
           end
         else
           it do
-            is_expected.to contain_package(package_name).with(
-              ensure:   'present',
-              require:  'Class[Zabbix::Repo]',
-              tag:      'zabbix'
-            )
+            is_expected.to contain_package(package_name).
+              with_ensure('present').
+              with_tag('zabbix').
+              that_requires('Class[zabbix::repo]')
           end
           it do
-            is_expected.to contain_service(service_name).with(
-              ensure:     'running',
-              enable:     true,
-              hasstatus:  true,
-              hasrestart: true,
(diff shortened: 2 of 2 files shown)


ekohl:
@bjschafer you added AIX support in https://github.com/voxpupuli/puppet-zabbix/pull/665. Could you verify I don't break anything with this?

manifests/agent.pp
```diff
@@ -432,24 +432,16 @@
...
-    require    => Package[$zabbix_package_agent],
+    ensure  => $service_ensure,
+    enable  => $service_enable,
+    require => Package[$zabbix_package_agent],
+  }
+
+  # Override the service provider on AIX
+  # Doing it this way allows overriding it on other platforms
+  if $facts['os']['name'] == 'AIX' {
+    Service[$servicename] { service_provider => 'init' }
```
igalic, line 444: given this documentation: https://puppet.com/docs/puppet/6.18/types/service.html#service-provider-init

```ruby
true == begin
      os = Facter.value(:operatingsystem).downcase
      family = Facter.value(:osfamily).downcase
      !(os == 'debian' || os == 'ubuntu' || family == 'redhat')
  end
```

i'm sceptical that puppet will know the correct `$service_path`, so maybe let's bring this back.

igalic (commented):
🙋🏻‍♀️

manifests/agent.pp
bastelfreak: I still dont get why we need to set the provider explicitly :(

ekohl: that's why I'm hoping @bjschafer can answer that.

igalic: because the default `service` provider is [src](https://puppet.com/docs/puppet/6.18/types/service.html#service-provider-src), but the zabbix package, i assume, provides an init style service file

ekohl: I did see that, but I don't know if `src` can run init style service files.

igalic:
do we have anything to fix the test failures? https://travis-ci.org/github/voxpupuli/puppet-zabbix/jobs/727040802

ekohl:
That was bad code from my side. I tried to properly use `that_requires` (which does proper checking in the catalog) instead of `with_require` (that only checks the property) but made a typo.

igalic (approved):

spec/classes/agent_spec.rb
```diff
@@ -68,24 +68,21 @@
...
-            )
+            is_expected.to contain_service(service_name).
+              with_ensure('running').
+              with_enable(true).
+              with_service_provider(facts[:osfamily] == 'AIX' ? 'init' : nil).
+              that_requires("Package[#{package_name}]")
           end
 
           it { is_expected.to contain_file(include_dir).with_ensure('directory') }
-          it { is_expected.to contain_zabbix__startup(service_name).with(require: "Package[#{package_name}]") }
```
ekohl, line 88: This `with_require` style happens more often in this module. I just fixed it for this file

igalic: let's open an issue as reminder to do it in a next iteration

igalic:
can we merge this?

spec/classes/agent_spec.rb
bastelfreak: what's bad about it?

bastelfreak (approved):

spec/classes/agent_spec.rb
ekohl: Consider this code:
```puppet
class myclass {
  package { 'mypackage':
  }

  file { '/etc/mypackage.conf':
    require => Package['mypackage'],
  }

  service { 'mypackage':
    subscribe => File['/etc/mypackage.conf'],
  }
}
```

Then the test code:
```ruby
it { is_expected.not_to contain_service('mypackage').with_require('Package[mypackage]') }
it { is_expected.to contain_service('mypackage').that_requires('Package[mypackage]') }
```

What I'm trying to show is that one tests for a parameter on a resource and the other the relationship exists in the catalog. That takes transitive requirements into account.

PR #712, opened 2020-09-11
Allow special characters in zabbixapi zabbix_pass & postgresql database_password

aclarkee:
This pull request fixes a couple of issues I came across regarding passwords with special characters not being handled correctly.

1. Zabbixapi not accepting passwords with special characters (I'm not too familiar with ruby, but this fix works in my tests)
2. postgresql echo exec not handling passwords with special characters correctly

### DIFF
--- a/lib/puppet/provider/zabbix.rb
+++ b/lib/puppet/provider/zabbix.rb
@@ -49,7 +49,7 @@ class Puppet::Provider::Zabbix < Puppet::Provider
     zbx = ZabbixApi.connect(
       url: "#{protocol}://#{api_config['default']['zabbix_url']}/api_jsonrpc.php",
       user: api_config['default']['zabbix_user'],
-      password: api_config['default']['zabbix_pass'],
+      password: "#{api_config['default']['zabbix_pass']}",
       http_user: api_config['default']['zabbix_user'],
       http_password: api_config['default']['zabbix_pass']
     )
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -84,7 +84,7 @@ class zabbix::database::postgresql (
   }
 
   exec { 'update_pgpass':
-    command => "echo ${database_host}:5432:${database_name}:${database_user}:${database_password} >> /root/.pgpass",
+    command => "echo '${database_host}:5432:${database_name}:${database_user}:${database_password}' >> /root/.pgpass",
     path    => "/bin:/usr/bin:/usr/local/sbin:/usr/local/bin:${database_path}",
     unless  => "grep \"${database_host}:5432:${database_name}:${database_user}:${database_password}\" /root/.pgpass",
     require => File['/root/.pgpass'],
--- a/spec/classes/database_postgresql_spec.rb
+++ b/spec/classes/database_postgresql_spec.rb
@@ -31,7 +31,7 @@ describe 'zabbix::database::postgresql' do
           end
 
           it { is_expected.to compile.with_all_deps }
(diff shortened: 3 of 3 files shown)


lib/puppet/provider/zabbix.rb
```diff
@@ -49,7 +49,7 @@ def self.create_connection
     zbx = ZabbixApi.connect(
       url: "#{protocol}://#{api_config['default']['zabbix_url']}/api_jsonrpc.php",
       user: api_config['default']['zabbix_user'],
```
bastelfreak, line 51: should we quote all of the attribute here, just to be future safe? Can you please add a comment to that explains why the quotes were added? I'm afraid otherwise they will be removed in the future

manifests/database/postgresql.pp
```diff
@@ -84,7 +84,7 @@
   }
 
   exec { 'update_pgpass':
-    command => "echo ${database_host}:5432:${database_name}:${database_user}:${database_password} >> /root/.pgpass",
+    command => "echo '${database_host}:5432:${database_name}:${database_user}:${database_password}' >> /root/.pgpass",
     path    => "/bin:/usr/bin:/usr/local/sbin:/usr/local/bin:${database_path}",
     unless  => "grep \"${database_host}:5432:${database_name}:${database_user}:${database_password}\" /root/.pgpass",
```
bastelfreak, line 89: this grep should be adjusted as well to add the single quotes to existing files?

spec/classes/database_postgresql_spec.rb
```diff
@@ -31,7 +31,7 @@
           end
 
           it { is_expected.to compile.with_all_deps }
-          it { is_expected.to contain_exec('update_pgpass').with_command('echo node01.example.com:5432:zabbix-server:zabbix-server:zabbix-server >> /root/.pgpass') }
+          it { is_expected.to contain_exec('update_pgpass').with_command('echo 'node01.example.com:5432:zabbix-server:zabbix-server:zabbix-server' >> /root/.pgpass') }
```
kenyon, line 34: This doesn't look right. Seems like there should be double quotation marks around the entire string, with single quotation marks within, like the lines below here.

lib/puppet/provider/zabbix.rb
kenyon: Also, could we add test(s) for usage of special characters?

spec/classes/database_postgresql_spec.rb
aclarkee: Thanks. I'm completely new to writing specs - so that was a quick attempt. I'll try your suggestion later today

lib/puppet/provider/zabbix.rb
aclarkee: I'll make the same change for `http_password` - since it uses the same password. I don't think there'd be any benefit for making this change for `user` and `http_user` though?

@kenyon Completely new to specs, and not very familiar with ruby.  I'll take a look at the existing specs and see if I can cobble something together. Would need a good review from someone more familiar with spec writing/ruby though.

Can I test specs locally, without having to push each change here for travis to test?

lib/puppet/provider/zabbix.rb
```diff
@@ -49,7 +49,7 @@ def self.create_connection
...
-      password: api_config['default']['zabbix_pass'],
+      password: "#{api_config['default']['zabbix_pass']}",
```
ekohl, line 52: I don't see why this would be needed. You're casting a string to another string.

ekohl:
In https://github.com/voxpupuli/puppet-zabbix/pull/904 I've removed the need for the pgpass file altogether.

PR #713, opened 2020-09-11
Zabbix5 agent: allowkey, denykey params

aclarkee:
Currently only works for zabbix5 when setting `zabbix::agent::zabbix_version: 5.0` through hiera. As I haven't messed with the zabbix_version logic. 

If someone can give some context on how zabbix_version should be used, I can give that a crack. I'm currently confused why this variable is being handled the way it is. (The Params class seems to just set the version based on OS, with default being 3.4?) Or whether there's a better way to get this information. A puppet fact maybe?

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -81,6 +81,17 @@
 #
 # [*enableremotecommands*]
 #   Whether remote commands from zabbix server are allowed.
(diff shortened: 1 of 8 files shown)


igalic (changes requested):
i'd like to see spec tests for this

PR #714, opened 2020-09-11
zabbix5: dbtls support

aclarkee:
Adds new zabbix server & proxy parameters for TLS connection to the database

Does not deploy the cafile - that is left as a manual task for now.

### DIFF
--- a/manifests/proxy.pp
+++ b/manifests/proxy.pp
@@ -108,6 +108,17 @@
 # [*database_port*]
 #   Database port when not using local socket. Ignored for sqlite.
 #
+# [*database_tlsconnect*]
+#   Setting this option enforces the use of TLS on connection to the database:
+#   required - connect using TLS
+#   verify_ca - connect using TLS and verify certificate
+#   verify_full - connect using TLS, verify certificate and verify that database identity specified by DBHost matches its certificate
+#   This parameter is supported since Zabbix 5.0.0.
+#
+# [*database_tlscafile*]
+#   Full pathname of a file containing the top-level CA(s) certificates for database certificate verification.
+#   This parameter is supported since Zabbix 5.0.0.
+#
 # [*localbuffer*]
 #   Proxy will keep data locally for N hours, even if the data have already been synced with the server
 #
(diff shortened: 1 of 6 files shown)


igalic (changes requested):
I'd like to see some spec tests for this

aclarkee:
> I'd like to see some spec tests for this

Could you give some guidance on spec test design, how to test them etc? I have zero experience in writing spec tests, unfortunately. That seems to be where most of the current pull requests are sitting at - I would like to contribute, but do not know how to contribute these test specs.

igalic:
a good way to start writing spec tests is by copying an existing test in the general vicinity of the parameters you're modifying, and go from there

we're also available to help you in #voxpupuli on freenode, or on the puppet slack

manifests/proxy.pp
```diff
@@ -108,6 +108,17 @@
 # [*database_port*]
```
igalic, line 108: the permissions on this file change
that's probably a mistake

manifests/server.pp
```diff
@@ -83,6 +83,17 @@
 # [*database_port*]
```
igalic, line 83: the permissions on this file change
that's probably a mistake

manifests/web.pp
```diff
@@ -215,6 +215,8 @@
   $database_password                                                  = $zabbix::params::server_database_password,
```
igalic, line 215: the permissions on this file change
that's probably a mistake

templates/zabbix_proxy.conf.erb
```diff
@@ -124,6 +124,8 @@ DBUser=<%= @database_user %>
 #	Comment this line if no password is used.
```
igalic, line 124: the permissions on this file change
that's probably a mistake

templates/zabbix_server.conf.erb
```diff
@@ -105,6 +105,8 @@ DBUser=<%= @database_user %>
 #	Comment this line if no password is used.
 #
```
igalic, line 106: the permissions on this file change
that's probably a mistake

igalic (changes requested):
👀

manifests/proxy.pp
aclarkee: Good pickup - thanks. I'll send through a commit to fix this later today.

PR #715, opened 2020-09-11
zabbix5: Allow DB creation

aclarkee:
Have only tested with postgresql. 

Duplicate of #694 - but there hasn't been any activity on that for 3 months. @bastelfreak Could you point me to the acceptance tests that need to be modified? (And is it possible to test these locally, or only through travis?)

On a side note: Looks like this logic can be simplified significantly if support for zabbix <3 is dropped. Looks like Zabbix themselves have dropped support for them, and only supports 3 onwards. Would you consider dropping support for zabbix versions <3?

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -28,9 +28,9 @@ class zabbix::database::postgresql (
   assert_private()
 
   #
-  # Adjustments for version 3.0/4.0 - structure of package with sqls differs from previous versions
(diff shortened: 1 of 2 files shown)


manifests/database/postgresql.pp
```diff
@@ -28,9 +28,9 @@
   assert_private()
```
igalic, line 28: this file changes permissions
that seems like a mistake

igalic (changes requested):

PR #720, opened 2020-10-19
Add parameters for SAML paths in zabbix.conf.php

wdberry:
Adds new parameters to allow specifying custom paths for the SAML certificates utilized for the SAML authentication configuration added in Zabbix 5.0.

This PR adds parameters for setting custom SAML certificate paths for the SAML authentication configuration introduced in Zabbix 5.0 based on the official documentation located here: https://www.zabbix.com/documentation/current/manual/web_interface/frontend_sections/administration/authentication#advanced_settings

n/a


manifests/init.pp
```diff
@@ -180,6 +180,9 @@
   String $additional_service_params         = $zabbix::params::additional_service_params,
   Optional[String[1]] $zabbix_user          = $zabbix::params::server_zabbix_user,
   Optional[String] $zabbix_server_name      = $zabbix::params::zabbix_server,
+  Optional[String] $sso_sp_cert             = $zabbix::params::sso_sp_cert,
```
bastelfreak, line 183: since the default is the same on every operating system, you can set the new parameters to default in the init.pp. You don't need to put it into the params.pp.
Also I think only absolute paths are allowed here, so we can harden the type even further:
```suggestion
  Optional[Stdlib::Absolutepath] $sso_sp_cert = undef,
```

wdberry: Because these parameters modify configuration specific to the Zabbix frontend web server, they're defined in both init.pp and web.pp, with the defaults in params.pp to have valid definitions if the top level class is included on a host (single host configuration) or if just the zabbix::web class is applied to the host (multiple host configuration).

They're also documented in web.pp, consistent with the rest of the parameters specific to web.pp.

You are correct that the only valid values are absolute paths, so typing them as Stdlib::Absolutepath would be the best choice. I initially considered typing them that way but with the lack of strong typing in the rest of the code I decided against it. Given that they're being typed at all, the best type should be used.

PR #721, opened 2020-10-19
Fix for apache config for Zabbix 5.0 and CentOS 7

wdberry:
Changes the apache::vhost configuration to work with the changes in Zabbix 5.0 regarding minimum PHP version and CentOS 7 by utilizing the up-to-date package names for the official packages.

This PR makes changes to support the updated frontend configuration required as detailed in the official documentation: https://www.zabbix.com/documentation/current/manual/installation/frontend/frontend_on_rhel7

The test portion is kind of sketchy without re-writing the entire spec to test for specific values of zabbix_version, however, will hopefully work correctly if the default zabbix_version changes.

n/a

### DIFF
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -369,7 +369,6 @@ class zabbix::web (
         require => Class['zabbix::repo'],
         tag     => 'zabbix',
       }
-
     }
   } # END case $facts['os']['name']
 
@@ -401,7 +400,7 @@ class zabbix::web (
   # Is set to true, it will create the apache vhost.
   if $manage_vhost {
     include apache
-    if facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '7') == 0 and versioncmp($zabbix_version, '5') >= 0 {
+    if $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '7') == 0 and versioncmp($zabbix_version, '5') >= 0 {
       include apache::mod::proxy
       include apache::mod::proxy_fcgi
       $apache_vhost_custom_fragment = ''
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -410,17 +410,17 @@ class zabbix::web (
         content => epp('zabbix/web/php-fpm.d.zabbix.conf.epp'),
       }
       $fcgi_filematch = {
-          path     => '/usr/share/zabbix',
-          provider => 'directory',
-          addhandlers => [
-            {
-              extensions => [
-                'php',
-                'phar',
-              ],
-              handler => 'proxy:unix:/var/opt/rh/rh-php72/run/php-fpm/zabbix.sock|fcgi://localhost',
-            },
-          ],
+        path     => '/usr/share/zabbix',
+        provider => 'directory',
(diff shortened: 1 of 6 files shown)


manifests/web.pp
```diff
@@ -337,6 +337,29 @@
         ],
       }
     }
+    'CentOS', 'RedHat': {
```
bastelfreak, line 340: why did you add RedHat here? Are you running it on RedHat enterprise?

templates/web/php-fpm.d.zabbix.conf.epp
```diff
@@ -0,0 +1,26 @@
+[zabbix]
```
bastelfreak, line 1: please add a header like `# THIS FILE IS MANAGED BY PUPPET`

manifests/web.pp
```diff
@@ -377,6 +401,47 @@
   # Is set to true, it will create the apache vhost.
   if $manage_vhost {
     include apache
+    if facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '7') == 0 and versioncmp($zabbix_version, '5') >= 0 {
```
bastelfreak, line 404: can you add an acceptance test for zabbix 5 to cover this? If you've questions about the tests or beaker, feel free to ping us on IRC in #voxpupuli on freenode.

manifests/web.pp
wdberry: Per the Zabbix documentation (https://www.zabbix.com/documentation/current/manual/installation/upgrade/packages/rhel_centos), the package changes apply to both CentOS 7 and RHEL 7 and due to the surrounding case statement evaluating $facts['os']['name'], I matched the pattern present in line 306 for defining the package to be the same on Ubuntu and Debian. That was the simplest way I could think of to include RHEL without refactoring the entire case statement to use a different fact.

Leaving out RedHat would resolve the issue I was making the change for on CentOS 7, but leave the issue present if someone was utilizing RHEL 7, because the official packages for CentOS 7 and RHEL 7 are the same.

templates/web/php-fpm.d.zabbix.conf.epp
wdberry: Will do.

manifests/web.pp
wdberry: Can do.

I have another pull request I was working on but hadn't submitted yet to resolve a specific issue related to a zabbix_proxy.conf parameter needing to be added to the template to allow overriding of the parameter to resolve an issue that occurred after upgrading to Zabbix 5 where the zabbix-proxy service would either fail to start correctly or start and then immediately terminate due to if I remember correctly selinux permissions.

I assume that not including those additional changes would cause a failure in the beaker tests for CentOS 7/RHEL 7 for a proxy node when selinux is enabled.

Should I merge those changes into this pull request, or keep it as a separate pull request and exclude that specific failing scenario from the acceptance tests?

wdberry: I think I answered my own question by reviewing the acceptance tests; I can update server_spec.rb within this PR and include an updated zabbix_proxy_spec.rb within the other PR.

PR #722, opened 2020-10-19
Add parameter for option StartPreprocessors

wdberry:
Adds parameters to set non-default values for option StartPreprocessors in zabbix_server.conf for Zabbix Server (supported since 3.4) and zabbix_proxy.conf for Zabbix Proxy (supported since 5.0)

I needed to be able to override this option in Foreman on my proxy due to the default being too low for my configuration.

n/a


manifests/params.pp
```diff
@@ -268,6 +268,7 @@
   $server_startsnmptrapper                  = '0'
   $server_starttimers                       = '1'
   $server_starttrappers                     = '5'
+  $server_startpreprocessors                = '3'
```
bastelfreak, line 271: I think only 1 to 1000 is allowed? In that case can you please use an integer and the `Integer[1,1000]` datatype?

wdberry: That would be correct.

The parameter is also not mandatory, so I'm going to switch it to Optional, with a value of undef, and add a conditional for the value in the template.

manifests/params.pp
```diff
@@ -268,6 +268,7 @@
   $server_startsnmptrapper                  = '0'
   $server_starttimers                       = '1'
   $server_starttrappers                     = '5'
+  $server_startpreprocessors                = undef
```
bastelfreak, line 271: if I read the change correctly, both new variables from params.pp are only used once. In that case please directly assign undef in init.pp / proxy.pp. That makes it later on easier to document the changes with puppet-strings.

PR #723, opened 2020-10-23
Add parameter and default for Option SocketDir

wdberry:
Fixes an issue that occurred after upgrading from Zabbix 4.0 to 5.0 that prevents the zabbix-proxy service from starting correctly.

Add parameter and default for Option SocketDir

Added the parameter socketdir to be able to override Option SocketDir in both zabbix_server.conf and zabbix_proxy.conf via an ENC (such as Foreman).

It was added in zabbix_server.conf for completeness, however, in Zabbix 5.0 the proxy service creates a Unix socket for preprocessing called zabbix_proxy_preprocessing.sock.

In the default location, /tmp, SELinux violations on creating the socket prevented the zabbix-proxy service from starting in my deployment on CentOS 7.

My workaround was to override SocketDir, which was added to zabbix_proxy.conf in 5.0, to change the directory that this socket is created in from the default to /var/run/zabbix.

Because it only exists when the service is running, /var/run seemed the appropriate place, and /var/run/zabbix already existed and had appropriate SELinux labels due to the agent and proxy PID files.

To resolve this issue without explicitly setting the parameter, I set the default for socketdir within zabbix::proxy to be /var/run/zabbix if zabbix_version is >= 5.0.

That seems a safe and sane enough default that works when both SELinux is enabled or disabled so that it's not necessary to do a boolean and comparison against both zabbix_version and facts('os.selinux.enabled').

n/a


templates/zabbix_server.conf.erb
```diff
@@ -516,3 +527,4 @@ LoadModulePath=<%= @loadmodulepath %>
 <% if @tlskeyfile %>TLSKeyFile=<%= @tlskeyfile %><% end %>
 
 <% end %>
+
```
kenyon, line 530: Remove extra newline. Would be good to also remove it from `zabbix_proxy.conf.erb`.
```suggestion
```

templates/zabbix_server.conf.erb
```diff
@@ -77,6 +77,17 @@ DebugLevel=<%= @debuglevel %>
...
 
+<% if @zabbix_version.to_f >= 3.4 %>
+### Option: SocketDir
+#       IPC socket directory.
+#               Directory to store IPC sockets used by internal Zabbix services.
+#
+# Mandatory: no
+# Default:
+
+SocketDir=<%= @socket_dir %>
```
kenyon, line 88:
```suggestion
SocketDir=<%= @socketdir %>
```

templates/zabbix_proxy.conf.erb
```diff
@@ -95,6 +95,17 @@ DebugLevel=<%= @debuglevel %>
...
 
+<% if @zabbix_version.to_f >= 5.0 %>
+### Option: SocketDir
+#       IPC socket directory.
+#               Directory to store IPC sockets used by internal Zabbix services.
+#
+# Mandatory: no
+# Default:
+
+SocketDir=<%= @socket_dir %>
```
kenyon, line 106:
```suggestion
SocketDir=<%= @socketdir %>
```

kenyon (changes requested):

templates/zabbix_server.conf.erb
wdberry: I evidently missed this in my internal refactoring, because I originally had the parameter as socket_dir rather than socketdir. Would also explain the failing tests!

templates/zabbix_server.conf.erb
wdberry: Should be done in both files in latest commit.

templates/zabbix_server.conf.erb
wdberry: Fixed in latest commit.

templates/zabbix_proxy.conf.erb
wdberry: Fixed in latest commit.

kenyon (approved):

bastelfreak (approved):

PR #726, opened 2020-10-29
Fix $hostinterfaceitem to not be an Fqdn.

### LINKED ISSUE #725: zabbix::agent::hostinterfaceitem not respected
Hello,

Setting `zabbix::agent::hostinterfaceitem`, e.g. to use DNS in host autoregistration, is not working correctly. I believe the problem is here:

https://github.com/voxpupuli/puppet-zabbix/blob/7ae8f5bd8d43cd279ffc0379db505bb3984e901f/manifests/agent.pp#L278

The type of `zabbix::agent::hostinterface` should indeed be `Optional[Stdlib::Fqdn]`, but `zabbix::agent::hostinterfaceitem` should be a string, since it expects a Zabbix item. See [the docs](https://www.zabbix.com/documentation/5.0/manual/appendix/config/zabbix_agentd) (just ctrl+f for "HostInterfaceItem").

bjschafer:
As per #725 and the [upstream documentation](https://www.zabbix.com/documentation/5.0/manual/appendix/config/zabbix_agentd), `zabbix::agent::hostinterfaceitem` should not be an FQDN as it refers to a Zabbix item.

I did not set a type, to match the similar `$hostmetadataitem`. Let me know if I should instead set it to be something like `Optional[String]`.

Fixes #725

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -275,7 +275,7 @@ class zabbix::agent (
   $hostmetadata                                   = $zabbix::params::agent_hostmetadata,
   $hostmetadataitem                               = $zabbix::params::agent_hostmetadataitem,
   Optional[Stdlib::Fqdn] $hostinterface           = $zabbix::params::agent_hostinterface,
-  Optional[Stdlib::Fqdn] $hostinterfaceitem       = $zabbix::params::agent_hostinterfaceitem,
+  $hostinterfaceitem                              = $zabbix::params::agent_hostinterfaceitem,
   $refreshactivechecks                            = $zabbix::params::agent_refreshactivechecks,
   $buffersend                                     = $zabbix::params::agent_buffersend,
   $buffersize                                     = $zabbix::params::agent_buffersize,


manifests/agent.pp
```diff
@@ -275,7 +275,7 @@
   $hostmetadata                                   = $zabbix::params::agent_hostmetadata,
   $hostmetadataitem                               = $zabbix::params::agent_hostmetadataitem,
   Optional[Stdlib::Fqdn] $hostinterface           = $zabbix::params::agent_hostinterface,
-  Optional[Stdlib::Fqdn] $hostinterfaceitem       = $zabbix::params::agent_hostinterfaceitem,
+  $hostinterfaceitem                              = $zabbix::params::agent_hostinterfaceitem,
```
bastelfreak, line 278: Please update this to:
```suggestion
  Optional[String[1]] $hostinterfaceitem          = $zabbix::params::agent_hostinterfaceitem,
```

bastelfreak:
Hi @bjschafer, thanks for the PR. please take a look at the inline comment I made and also rebase against our latest master branch. can you please add a test that checks that it passes with a string that's not an fdqn? Otherwise I'm afraid in the future people might add the fqdn type again.

bastelfreak:
@bjschafer can you please rebase against our current master branch and squash the commits?

Valantin:
hi @bjschafer can you update this PR with information provided in #953 otherwise I can reimplement it in a new PR

PR #728, opened 2020-11-25
modulesync 4.0.0

bastelfreak:
modulesync 4.0.0


ekohl (changes requested):
EL6 looks broken (previously didn't run). Recommending to drop EL6 here.

Other test are still running, but looks like this needs `pidfile_workaround` on Ubuntu and CentOS.

Issue #689, opened 2020-05-14
Add Support for Zabbix 5.0


broadaxe:
The current version of this module does not support Zabbix 5.0. I know this Zabbix version just got released a couple of days ago, but it would be awesome if/when it is supported. I understand this would take time.
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: Latest
- Ruby: Latest with Puppet
- Distribution: CentOS 7
- Module version: Latest

## How to reproduce (e.g Puppet code you use)
Use the current version of these Puppet library to deploy a Zabbix 5.0 server.
## What are you seeing
Some of the packages have been renamed, so it fails to install the old names. In the end, the web based UI for Zabbix is broken.

## What behaviour did you expect instead
I would guess I was to optimistic, I expected to have a working Zabbix server.

VibroAxe:
I would also be keen to see this

msnellen:
I'm also very interested in this as we are currently managing only our Zabbix-agent and would like to migrate to Zabbix 5 in the near future. Would be great if we could use this module to handle the whole thing.

aclarkee:
I believe the main blocker here will be that the ZabbixAPI rubygem only officially supports up to 4.2, and hasn't had any update since [November 2019](https://github.com/express42/zabbixapi)

I wonder where that leaves this module, since it relies heavily on this rubygem.. Either branching off and maintaining a new version of the gem, or talking to the API directly?..

jameskirsop:
At this point, I'm probably going to have to abandon using this module because of the need to move to 5.0 rapidly.

I'm not using the features of this module that require the ZabbixAPI rubygem - and if this is a roadblock to getting 5.0 support, I'd be all for removing support for automatically adding hosts with agents via the API.

aclarkee:
I've actually been doing some testing with zabbix 5.0 - the zabbixapi gem works fine, just warns. 

I'll submit a pull request a bit later, but my usecase is very specific, so not everything is tested. Someone else will have to step up and contribute code to get their usecases working :)

jameskirsop:
@aclarkee Do you have a branch for your changes that I can take a look at and test with (and possibly contribute to)?

aredoff:
We really need zabbix_agent2 support!
Please add new agent support to your wonderful module.

aclarkee:
@jameskirsop Sorry for the late reply, got sidetracked with other priorities. I've just submitted a couple of pull requests with my modifications.

[...]

PR #736, opened 2021-01-21
Better CentOS7 / Zabbix5.0 support

jameskirsop:
This PR more fully addresses #689 by including the 'frontend' repo that's required for CentOS7 to include PHP72. Also ensures the SCL repo is present and installs the appropriate Zabbix Web package instead of the previous (and current) generic one.

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -92,6 +92,22 @@ class zabbix::repo (
           gpgkey   => $gpgkey_nonsupported,
           priority => '1',
         }
+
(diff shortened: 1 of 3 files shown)


jameskirsop:
Can see this is now failing for the OS targeted by the PR, but I'm not sure how/where to update the tests - or how to tell why they're failing (the debugging is a bit confusing to me).

bastelfreak:
Hi @jameskirsop . can you hop into #voxpupuli on freenode IRC or http://slack.puppet.com/ ? we can guide you through the tests and how to update them.

PR #737, opened 2021-01-21
Adds 5.x in the version check for which schema path to use

jameskirsop:
Helps address #689 by allowing for the correct database schema path to be used when doing the initial installation for a 5.0 install.

Also just realised that this is a dupe of both #715 and #694 and we're in the same holding pattern now because of the acceptance tests. Maybe a maintainer (@bastelfreak?) can assist in pointing @aclarkee, @stanhva and myself in the right direction?

### DIFF
--- a/manifests/database/mysql.pp
+++ b/manifests/database/mysql.pp
@@ -28,9 +28,9 @@ class zabbix::database::mysql (
   assert_private()
 
   #
-  # Adjustments for version 3.0/4.0 - structure of package with sqls differs from previous versions
+  # Adjustments for version 3.0 and onwards - structure of package with sqls differs from previous versions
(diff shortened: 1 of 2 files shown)


bastelfreak:
hey!
so the acceptance tests are in https://github.com/voxpupuli/puppet-zabbix/tree/master/spec/acceptance. that's basically puppet code that gets executed in a docker container. some of the tests install the zabbix server in a specific version. It would be good if at least one of them gets extended (simply copy and paste the puppet code + tests and append it). https://github.com/voxpupuli/puppet-zabbix/blob/master/spec/acceptance/server_spec.rb in particular would be a good candidate. it installs a minimal zabbix server and afterwards checks if the service + database is running.

PR #739, opened 2021-02-05
Add missing parameters for logfile & sslcertlocation & sslkeylocation

shaunbugler:
#### zabbix::agent
- logtype param was not defined in zabbix::params, also had different type enforcement that other classes

#### zabbix::server
- logfile param had different type enforcement that other classes

#### zabbix::proxy
- proxy config caters for sslcertlocation and sslkeylocation but these were not passable as params
  - also standardized the naming to match other modules in template
- logtype had different type enforcement that other classes
- fixed template to not put done extra blank lines midconfig


manifests/params.pp
```diff
@@ -431,6 +431,17 @@
     default => '/var/run/zabbix',
   }
 
+  # provided by camp2camp/systemd
```
kenyon, line 434:
```suggestion
  # provided by camptocamp/systemd
```

kenyon (changes requested):
It seems there are several unrelated changes that should be in separate pull requests:

1. `$logtype` and `$proxy_log*` change for agent class.
2. Addition of `$proxy_ssl*` parameters for proxy class.
3. Template spacing.

There should also be tests added for the new `$proxy_log*` variable logic.

PR #742, opened 2021-02-28
Drop Puppet 5 support

bastelfreak:

Issue #734, opened 2021-01-19
Need to set $DB[DOUBLE_IEEE754] to true in Zabbix >=5.0's zabbix.conf.php.


Joshua-Snapp:
## How to reproduce (e.g Puppet code you use)
Install or upgrade to Zabbix 5.0 or newer.
## What are you seeing
The frontend will display "Database history tables upgraded: No" in the System Information widget in the Global view dashboard, unless `$DB['DOUBLE_IEEE754']` value is set to true in `/ui/conf/zabbix.conf.php`.
## Any additional information you'd like to impart
This is discussed in the [ENABLING EXTENDED RANGE OF NUMERIC (FLOAT) VALUES](https://www.zabbix.com/documentation/5.0/manual/installation/upgrade_notes_500#enabling_extended_range_of_numeric_float_values) section of the upgrade notes for 5.0.

Issue #693, opened 2020-06-09
Allow for configuration of 5.0's zabbix.conf.php SAML settings


jameskirsop:
5.0 introduced some new settings that are managed via `zabbix.conf.php` including those for SAML authentication. On a CentOS 7 installation this file is located in `/etc/zabbix/web`.

These would be added to the `zabbix::web` class.

Particular configuration settings to manage:
```
$SSO['SP_KEY'] = '<path to the SP private key file>';
$SSO['SP_CERT'] = '<path to the SP cert file>';
$SSO['IDP_CERT'] = '<path to the IDP cert file>';
$SSO['SETTINGS']
```

Reference: https://www.zabbix.com/documentation/current/manual/web_interface/frontend_sections/administration/authentication#saml_authentication

anthonysomerset:
there is also a DB config value relating to the DB in the 5.0 upgrade 

https://www.zabbix.com/documentation/current/manual/installation/upgrade_notes_500 - Enabling extended range of numeric (float) values

PR #743, opened 2021-02-28
Add support for Zabbix 5.x

### LINKED ISSUE #731: the install action for zabbix 5.0 on centos/redhat 7 is wrong

- Puppet: puppet5-release-5.0.0-7
- Ruby: 
- Distribution: CentOS Linux release 7.7.1908
- Module version: 8.0.0

class { 'zabbix':
      zabbix_version  => $version,
}
set $version = 5.0
## What are you seeing
puppet zabbix module try to install zabbix-web-mysql
according to zabbix official install instruction, only on centos/redhat 8 can find the zabbix-web-mysql package. when it is centos7, you need to install centos-release-scl and then install zabbix-web-mysql-scl package from the scl repo.
## Output log
Error: Could not update: Execution of '/usr/bin/yum -d 0 -e 0 -y install zabbix-web-mysql' returned 1: Error: Nothing to do
Error: /Stage[main]/Zabbix::Web/Package[zabbix-web-mysql]/ensure: change from 'purged' to 'latest' failed: Could not update: Execution of '/usr/bin/yum -d 0 -e 0 -y install zabbix-web-mysql' returned 1: Error: Nothing to do (corrective)

root-expert:
Adds support for Zabbix 5.x.
Maybe more tests are needed
Depends on #742 

Fixes #731 
Closes #737 
Closes #736 
Closes #721 
Closes #715 
Closes #694

### DIFF
--- a/manifests/database/mysql.pp
+++ b/manifests/database/mysql.pp
@@ -30,7 +30,7 @@ class zabbix::database::mysql (
   #
   # Adjustments for version 3.x/4.x/5.x - structure of package with sqls differs from previous versions
   case $zabbix_version {
-    /^(3|4|5).\d+$/: {
+    /^[345].\d+$/: {
       if ($database_schema_path == false) or ($database_schema_path == '') {
         $schema_path   = '/usr/share/doc/zabbix-*-mysql*'
       }
--- a/manifests/database/mysql.pp
+++ b/manifests/database/mysql.pp
@@ -28,9 +28,9 @@ class zabbix::database::mysql (
   assert_private()
 
   #
-  # Adjustments for version 3.0/4.0 - structure of package with sqls differs from previous versions
+  # Adjustments for version 3.x/4.x/5.x - structure of package with sqls differs from previous versions
   case $zabbix_version {
-    /^(3|4).\d+$/: {
+    /^(3|4|5).\d+$/: {
       if ($database_schema_path == false) or ($database_schema_path == '') {
         $schema_path   = '/usr/share/doc/zabbix-*-mysql*'
       }
--- a/spec/acceptance/zabbix_proxy_spec.rb
+++ b/spec/acceptance/zabbix_proxy_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper_acceptance'
 require 'serverspec_type_zabbixapi'
 
 # rubocop:disable RSpec/LetBeforeExamples
-describe 'zabbix_proxy type', unless: default[:platform] =~ %r{debian-10-amd64} do
+describe 'zabbix_proxy type', unless: default[:platform] =~ %r{(ubuntu-16.04|debian-9|debian-10)-amd64} do
   context 'create zabbix_proxy resources' do
     # This will deploy a running Zabbix setup (server, web, db) which we can
     # use for custom type tests
@@ -20,7 +20,7 @@ describe 'zabbix_proxy type', unless: default[:platform] =~ %r{debian-10-amd64} 
       -> class { 'postgresql::server': }
 
       class { 'zabbix':
-        zabbix_version   => '4.4',
+        zabbix_version   => '5.0',
         zabbix_url       => 'localhost',
         zabbix_api_user  => 'Admin',
         zabbix_api_pass  => 'zabbix',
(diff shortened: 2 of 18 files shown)


DJClean (commented):
As I was looking into upgrading to Zabbix 5.0 myself, looked into this PR

But it seems to break on web.pp because of the custom_fragment with all the php_value settings
The var $apache_vhost_custom_fragment is being set, but not used to set the fragment in the vhost if I saw it right. Overall it seems to work just fine if I remove the fragment manually

root-expert:
Thanks! I'll look into it and report back

DJClean:
Already pushed a PR to your branch with the fix :)

spec/acceptance/zabbix_template_host_spec.rb
```diff
@@ -15,7 +15,7 @@ class { 'apache':
         include postgresql::server
 
         class { 'zabbix':
-          zabbix_version   => '4.4',
```
bastelfreak, line 18: since we still support zabbix 4.4 in the module: can you copy the whole test and use `5.0` is the second block? so we test both versions

root-expert: Do you think it's wise to test for the most versions in a do loop? Like i did in some specs. Probably from 4.0 and upwards?

bastelfreak: that would even be better. but I don't know if they all will pass on the first run. If you like you can give it a try, otherwise that might be a good thing for another PR.

manifests/database/mysql.pp
```diff
@@ -28,9 +28,9 @@
   assert_private()
 
   #
-  # Adjustments for version 3.0/4.0 - structure of package with sqls differs from previous versions
+  # Adjustments for version 3.x/4.x/5.x - structure of package with sqls differs from previous versions
   case $zabbix_version {
-    /^(3|4).\d+$/: {
+    /^[345].\d+$/: {
```
bastelfreak, line 33: if you like you could also just remove the case statement. we really don't support zabbix 1/2 anymore :D

root-expert: True :D

bastelfreak: mhm since some tests still run on 2.x, this might be something for a cleanup PR.

root-expert: I mean if i remove all the legacy code the PR could get pretty big

jameskirsop:
Could we also consider rolling in [my branch for adding support for the SAML / SSO settings](https://github.com/jameskirsop/puppet-zabbix/tree/add-saml-config-support) introduced in 5.x? Doing this will allow #693 to be closed.

If it's deemed not appropriate I'll file a seperate PR.

jadestorm:
Not to pile on  =)  But this setting is pretty important for Zabbix 5:
https://github.com/voxpupuli/puppet-zabbix/issues/734
(it has a PR already)

jadestorm:
PS aside from that I am using @root-expert 's PR fork directly right now for my 5.2 install and it's working great!  Mind you I'm not (yet) using the exported resources functionality.

root-expert:
> Could we also consider rolling in [my branch for adding support for the SAML / SSO settings](https://github.com/jameskirsop/puppet-zabbix/tree/add-saml-config-support) introduced in 5.x? Doing this will allow #693 to be closed.
> 
> If it's deemed not appropriate I'll file a seperate PR.

I think you should open a new request after rebasing with my branch. Also you should write some specs ;) 

> Not to pile on =) But this setting is pretty important for Zabbix 5:
> #734
> (it has a PR already)

I believe that PR could be merged if rebased as well.

spec/acceptance/zabbix_proxy_spec.rb
```diff
@@ -2,45 +2,49 @@
...
+      it 'works with no error on first apply' do
+        # Cleanup old database
+        prepare_host
 
-    it 'works idempotently with no errors' do
-      # Run it twice and test for idempotency
-      apply_manifest(pp1, catch_failures: true)
-      apply_manifest(pp1, catch_changes: true)
-    end
+        apply_manifest(pp1, catch_failures: true)
```
root-expert, line 40: I don't know why though it was idempotent before. 
It's weird because we are running the same manifests on the other specs (application, host etc).
Apache module isn't idempotent.

bastelfreak (approved):

spec/acceptance/zabbix_proxy_spec.rb
bastelfreak: apache being apache ¯\_(ツ)_/¯. I think we can ignore this since it was idempotent before as well.

kenyon (approved):

PR #744, opened 2021-03-03
postgres: Use UTF8 encoding and postgres 12 in acceptance tests

bastelfreak:


bastelfreak:
the zabbix DB contains very specific encoded strings, without this the tests, on centos 7, won't pass.

bastelfreak:
also for strange reasons, postgres 9.5 in ubuntu 16 fails in docker.

PR #746, opened 2021-03-06
acceptance tests: move package setup into helper

bastelfreak:
We rely on those packages in most tests. To be able to run them
independently, we need to manage them in the helper.


bastelfreak:
this PR also contains https://github.com/voxpupuli/puppet-zabbix/pull/744

PR #748, opened 2021-03-07
database: use namespaced postgresql_password function; bump puppetlabs/postgresql lower dependency 4.4.2->6.4.0

kenyon:
Fixes deprecation warning.

Also increased minimum postgresql module version to when namespaced postgresql_password function was introduced.

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -149,7 +149,7 @@ class zabbix::database (
         postgresql::server::db { $database_name:
           user       => $database_user,
           owner      => $database_user,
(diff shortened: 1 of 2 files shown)


bastelfreak:
mhm we need to check which puppetlabs-postgresql version introduced this and maybe bumo thr dependency boundary in metadata.json

kenyon:
> mhm we need to check which puppetlabs-postgresql version introduced this and maybe bumo thr dependency boundary in metadata.json

Forgot about that. Done.

PR #750, opened 2021-03-14
Drop Zabbix 2/3 support and deprecated parameters; only support zabbix 4.0, 5.0, 5.2

bastelfreak:
Only Zabbix 4 and newer are supported upstream. This PR drops support
for older releases. See https://www.zabbix.com/de/life_cycle_and_release_policy
for current versions that upstream supports.


spec/classes/repo_spec.rb
```diff
@@ -169,115 +100,74 @@
...
-                manage_repo: true
-              }
-            end
-
-            it { is_expected.to contain_yumrepo('zabbix').with_baseurl('https://repo.zabbix.com/zabbix/2.2/rhel/6/$basearch/') }
-            it { is_expected.to contain_yumrepo('zabbix').with_gpgkey('https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX') }
-            it { is_expected.to contain_yumrepo('zabbix-nonsupported').with_baseurl('https://repo.zabbix.com/non-supported/rhel/6/$basearch/') }
-            it { is_expected.to contain_yumrepo('zabbix-nonsupported').with_gpgkey('https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX') }
-          end
+        when '8'
```
root-expert, line 103: We don't support RedHat 8

spec/classes/repo_spec.rb
```diff
@@ -169,115 +100,74 @@
...
-            it { is_expected.to contain_yumrepo('zabbix-nonsupported').with_baseurl('https://repo.zabbix.com/non-supported/rhel/6/$basearch/') }
-            it { is_expected.to contain_yumrepo('zabbix-nonsupported').with_gpgkey('https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX') }
+            it { is_expected.to contain_yumrepo('zabbix').with_baseurl('https://repo.zabbix.com/zabbix/5.0/rhel/8/$basearch/') }
+            it { is_expected.to contain_yumrepo('zabbix').with_gpgkey('https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-A14FE591') }
+            it { is_expected.to contain_yumrepo('zabbix-nonsupported').with_baseurl('https://repo.zabbix.com/non-supported/rhel/8/$basearch/') }
+            it { is_expected.to contain_yumrepo('zabbix-nonsupported').with_gpgkey('https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-79EA5ED4') }
           end
 
-          context 'on RedHat 6 and Zabbix 3.2' do
+          context 'on RedHat 8 and Zabbix 4.0' do
```
root-expert, line 133: Same here

spec/classes/repo_spec.rb
bastelfreak: cleaned it up

spec/classes/repo_spec.rb
bastelfreak: cleaned it up

root-expert (approved):

bastelfreak:
This PR also contains https://github.com/voxpupuli/puppet-zabbix/pull/751

README.md
```diff
@@ -504,47 +504,9 @@ There are some zabbix specific parameters, please check them by opening the mani
...
-  * OracleLinux 5.x, 6.x
-  * Scientific Linux 5.x, 6.x
-  * Ubuntu 12.04
-  * Debian 6, 7
-  * xenserver 6
+
+This module supports Zabbix 4.0 to 5.2. The upstream supported versions are documented [here](https://www.zabbix.com/de/life_cycle_and_release_policy)
+Please have a look into the metadata.json for all supported operating systems.
 
 This module is supported on both the community as the Enterprise version of Puppet.
```
ekohl, line 511: Unrelated, but 
```suggestion
This module is supported on both the community and the Enterprise version of Puppet.
```

manifests/agent.pp
```diff
@@ -321,20 +321,12 @@
...
   if $zabbix_version != $zabbix::params::zabbix_version and $service_type == $zabbix::params::service_type {
-    $real_service_type = versioncmp($zabbix_version, '3.0') ? {
-      1  => 'simple',
-      0  => 'simple',
-      -1 => 'forking',
-    }
+    $real_service_type = 'simple'
   } else {
     $real_service_type = $service_type
   }
```
ekohl, line 332: I don't understand this logic. If there is no differentiation on version anymore, can't you just trust `$service_type`? Same above with `additional_service_params`.

manifests/server.pp
```diff
@@ -405,11 +382,7 @@
   # but only if the variable isn't provided.
 
   if $zabbix_version != $zabbix::params::zabbix_version and $additional_service_params == $zabbix::params::additional_service_params {
```
ekohl, line 384: Same as earlier: can't you just trust the param now?

manifests/web.pp
```diff
@@ -275,24 +275,9 @@
...
-      '3.2' : {
-        $zabbixapi_version = '3.2.1'
-      }
-      '3.4' : {
-        $zabbixapi_version = '4.0.0'
-      }
       '4.0': {
         $zabbixapi_version = '4.2.0'
       }
-      '4.4': {
```
ekohl, line 293: This suggests 4.4 is no longer supported but 4.0 is. Is that correct?

manifests/agent.pp
bastelfreak: ah good  catch. I cleaned that up

manifests/web.pp
bastelfreak: yep. 4.0 is the supported LTS. additionally supported are 5.0 and 5.2

ekohl (commented):
Haven't done a very thorough review but I don't see more issues. The only thing is that I suspect that if previously the user set version to 4.4, that no longer works for web. The PR title doesn't make that clear which can surprise users.

root-expert:
Maybe we should update the README with all the supported versions?

Fabian1976:
Is it smart to restrict versions like that?
This module isn't know for its quick releases and patches (last release was more then 1 year ago). Zabbix release 5.4 is already out and will probably work fine with this module, except for the strict version lock which prevents it from running. I expect Zabbix to drop 5.2 support pretty soon (as they always do when a new, non LTS, version is released).

I hope the release cycle for this module will be sped up...

PR #751, opened 2021-03-14
Bump default Zabbix version on linux 3.4->5.0

bastelfreak:
3.4 is EOL and we will remove code for it in https://github.com/voxpupuli/puppet-zabbix/pull/750

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -138,10 +138,8 @@ class zabbix::params {
 
   if downcase($facts['kernel']) == 'windows' {
     $zabbix_version = '4.4.5'
-  } elsif $facts['os']['name'] == 'Debian' and Integer($facts['os']['release']['major']) == 10 {
-    $zabbix_version =  '4.0'
   } else {
-    $zabbix_version = '3.4'
+    $zabbix_version = '5.0'
   }
 
   $manage_startup_script = downcase($facts['kernel']) ? {
--- a/spec/classes/repo_spec.rb
+++ b/spec/classes/repo_spec.rb
@@ -28,16 +28,16 @@ describe 'zabbix::repo' do
         end
 
         case facts[:os]['release']['major']
-        when '6'
-          context 'on Debian 6 and Zabbix 2.0' do
+        when '10'
+          context 'on Debian 10 and Zabbix 4.0' do
             let :params do
               {
-                zabbix_version: '2.0',
+                zabbix_version: '4.0',
                 manage_repo: true
               }
             end
 
-            it { is_expected.to contain_apt__source('zabbix').with_location('http://repo.zabbix.com/zabbix/2.0/debian/') }
+            it { is_expected.to contain_apt__source('zabbix').with_location('http://repo.zabbix.com/zabbix/4.0/debian/') }
           end
 
         when '7'
--- a/spec/classes/web_spec.rb
+++ b/spec/classes/web_spec.rb
@@ -13,17 +13,18 @@ describe 'zabbix::web' do
   end
 
   on_supported_os.each do |os, facts|
-    next if facts[:os]['name'] == 'windows'
-    context "on #{os} " do
-      let :facts do
-        facts
-      end
+    %w[4.0 5.0].each do |zabbix_version|
+      next if facts[:os]['name'] == 'windows'
+      next if facts[:os]['name'] == 'Archlinux'
+      next if facts[:os]['name'] == 'Gentoo'
+      next if facts[:os]['name'] == 'Debian' and facts[:os]['release']['major'] == '9'
+      next if facts[:os]['name'] == 'Ubuntu' and facts[:os]['release']['major'] == '16.04'
 
-      if facts[:osfamily] == 'Archlinux' || facts[:osfamily] == 'Gentoo'
-        context 'with all defaults' do
-          it { is_expected.not_to compile }
+      context "on #{os} " do
+        let :facts do
+          facts
         end
-      else
+
         context 'with all defaults' do
           it { is_expected.to compile.with_all_deps }
           it { is_expected.to contain_file('/etc/zabbix/web').with_ensure('directory') }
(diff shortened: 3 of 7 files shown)


spec/classes/web_spec.rb
```diff
@@ -53,7 +55,7 @@
           it { is_expected.not_to contain_selboolean('httpd_can_connect_zabbix') }
         end
 
-        %w[4.4 5.0].each do |zabbix_version|
+        %w[4.0 5.0].each do |zabbix_version|
           describe "with database_type as postgresql and zabbix_version #{zabbix_version}" do
             next if zabbix_version == '5.0' and %w[Debian Ubuntu].include?(facts[:osfamily]) and
```
root-expert, line 60: I think I'm doing the same here as you are doing at beginning of the file. Maybe this can be removed then?

root-expert: https://github.com/voxpupuli/puppet-zabbix/pull/751/files/fa5f23d200347590a247724b01850f24908f63e4#diff-82ccb785b8e760dc28ab7067a48ebd387c93d71c401fbebb9256bfa38117014aR17

bastelfreak: cleaned it up

spec/classes/web_spec.rb
```diff
@@ -201,7 +198,8 @@
               database_password: 'zabbix-server',
               zabbix_server: 'localhost',
               zabbix_listenport: '3306',
-              zabbix_server_name: 'localhost'
+              zabbix_server_name: 'localhost',
+              zabbix_version: '4.0'
```
root-expert, line 202: Maybe we should use #{zabbix_version} here as well. No big deal though.

root-expert (approved):

spec/classes/repo_spec.rb
```diff
@@ -27,17 +27,17 @@
           it { is_expected.to contain_apt__source('zabbix').with_location('https://example.com/foo') }
         end
 
-        case facts[:os]['release']['major']
-        when '6'
-          context 'on Debian 6 and Zabbix 2.0' do
+        case facts[:os]['release']['major'].to_i
+        when 10
```
ekohl, line 31: I would avoid `.to_i`. It's always a string and your code doesn't become easier with integers
```suggestion
        case facts[:os]['release']['major']
        when '10'
```

You also introduce a subtle bug on line 43 which now no longer matches.

spec/classes/web_spec.rb
```diff
@@ -13,17 +13,18 @@
...
-      end
-
-      if facts[:osfamily] == 'Archlinux' || facts[:osfamily] == 'Gentoo'
-        context 'with all defaults' do
-          it { is_expected.not_to compile }
+    %w[4.0 5.0].each do |zabbix_version|
+      next if facts[:os]['name'] == 'windows'
+      next if facts[:os]['name'] == 'Archlinux'
+      next if facts[:os]['name'] == 'Gentoo'
+      next if facts[:os]['name'] == 'Debian' and facts[:os]['release']['major'].to_i == 9
```
ekohl, line 20: I'd also avoid `.to_i` here
```suggestion
      next if facts[:os]['name'] == 'Debian' and facts[:os]['release']['major'] == '9'
```

spec/classes/web_spec.rb
```diff
@@ -53,52 +54,48 @@
...
-            end
-            it { is_expected.to contain_file('/etc/zabbix/web/zabbix.conf.php').with_content(%r{^\$DB\['TYPE'\]     = 'POSTGRESQL'}) }
+          packages.each do |package|
+            it { is_expected.to contain_package(package) }
           end
+          it { is_expected.to contain_file('/etc/zabbix/web/zabbix.conf.php').with_content(%r{^\$DB\['TYPE'\]     = 'POSTGRESQL'}) }
         end
 
-        describe 'with database_type as mysql' do
+        describe 'with database_type as mysql', unless: facts[:os]['release']['major'].to_i == 7 do
```
ekohl, line 98: And here as well
```suggestion
        describe 'with database_type as mysql', unless: facts[:os]['release']['major'] == '7' do
```

ekohl (dismissed):

manifests/params.pp
```diff
@@ -138,10 +138,8 @@
 
   if downcase($facts['kernel']) == 'windows' {
     $zabbix_version = '4.4.5'
-  } elsif $facts['os']['name'] == 'Debian' and Integer($facts['os']['release']['major']) == 10 {
-    $zabbix_version =  '4.0'
```
ekohl, line 142: I dislike this because Debian 10 ships version 4.0 in the repo. So out of the box it would be wrong. Also using the external repository is not set by default. IMHO it should always prefer distro default versions, so that part is correct but the combination of the two isn't good.

root-expert: manage_repo is set be default to true. So it will always try to find the specified version from repo.zabbix.com.
I think changing it to 5.0  by default is okay for Debian as well.

ekohl: Hmm, personally I hate it when modules do that but since I don't use this module I guess I don't care too much.

bastelfreak: @ekohl manage_repo is true by default for Debian/Ubuntu: https://github.com/voxpupuli/puppet-zabbix/blob/master/manifests/params.pp#L16-L21
There were no zabbix packages for the old default version and we didn't want to bump that for every OS to avoid a backwards incompatible release. that's why we set it to zabbix 4 on debian 10 while implementing that.

ekohl: It was confusing because the default version on Debian 10 is also 4: https://packages.debian.org/buster/zabbix-agent (and other zabbix-* packages have matching versions). That's why I assumed no external repo was used.

bastelfreak: @ekohl can you approve this or is there anything left that should be changed?

spec/classes/web_spec.rb
```diff
@@ -53,52 +54,48 @@
...
-            end
-            it { is_expected.to contain_file('/etc/zabbix/web/zabbix.conf.php').with_content(%r{^\$DB\['TYPE'\]     = 'POSTGRESQL'}) }
+          packages.each do |package|
+            it { is_expected.to contain_package(package) }
           end
+          it { is_expected.to contain_file('/etc/zabbix/web/zabbix.conf.php').with_content(%r{^\$DB\['TYPE'\]     = 'POSTGRESQL'}) }
         end
 
-        describe 'with database_type as mysql' do
+        describe 'with database_type as mysql', unless: facts[:os]['release']['major'] == '7' do
```
ekohl, line 98: I'm generally not a fan of checking a major release without the OS family/name,

bastelfreak: will update it

bastelfreak: done

PR #755, opened 2021-04-27
Drop EoL Ubuntu 16.04; Add Ubuntu 18.04

bastelfreak:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -101,7 +101,7 @@
     {
       "operatingsystem": "Ubuntu",
       "operatingsystemrelease": [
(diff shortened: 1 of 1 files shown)


root-expert:
If you need some help --> https://github.com/root-expert/puppet-zabbix/commit/a1195fd4b6a86424536221238f3b5fd83b1df1b8

Still some test on 5.2 are failing, maybe it's an Ubuntu problem not sure yet :man_shrugging:

bastelfreak:
thanks for your fix!

PR #756, opened 2021-05-03
Fix templates importing in Zabbix 5.2+

root-expert:
Zabbix changed templateScreens option in configurations.import API call to templateDashboards.

Source: 

5.2: https://www.zabbix.com/documentation/current/manual/api/reference/configuration/import
5.0: https://www.zabbix.com/documentation/5.0/manual/api/reference/configuration/import
(rules object)

### DIFF
--- a/manifests/template.pp
+++ b/manifests/template.pp
@@ -12,6 +12,9 @@
 # [*templ_source*]
 #  The location of the XML file wich needs to be imported.
 #
+# [*zabbix_version*]
+#  The Zabbix version on which the template will be installed on.
+#
 # === Example
 #
 #  zabbix::template { 'Template App MySQL':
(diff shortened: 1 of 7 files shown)


manifests/template.pp
```diff
@@ -27,11 +27,13 @@
 # Copyright 2015  Vladislav Tkatchev
 #
 define zabbix::template (
-  $templ_name   = $title,
-  $templ_source = '',
+  $templ_name     = $title,
+  $templ_source   = '',
+  $zabbix_version = $zabbix::params::zabbix_version,
```
bastelfreak, line 32: thanks for the PR! Can you pleae add a datatype here? I guess a simple string is okay:
```suggestion
  String[1] $zabbix_version = $zabbix::params::zabbix_version,
```

bastelfreak: please also add some documentation for the new parameter in the header.

root-expert: Sure!

root-expert: Done. You can cancel some runs if you want to btw :stuck_out_tongue:

PR #757, opened 2021-05-03
Add Ubuntu 18.04 and 20.04 support

root-expert:
This needs #756 to pass

Partially contains #755


root-expert:
Rebased with master

PR #762, opened 2021-05-03
modulesync 4.1.0

bastelfreak:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,8 @@ jobs:
     needs: setup_matrix
     runs-on: ubuntu-latest
     env:
+      LANG: en_US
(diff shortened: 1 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -71,7 +71,5 @@ jobs:
       - name: Run tests
         run: bundle exec rake beaker
         env:
-          LANG: en_US
-          LC_ALL: en_US.UTF-8
```
root-expert, line 75: Wouldn't that break Puppet 7 runs on CentOS 7?

bastelfreak: I want to see if they fail. If that's the case I add it back

bjoernhaeuser (approved):

PR #763, opened 2021-05-05
Implement additional SAML settings

### LINKED ISSUE #693: Allow for configuration of 5.0's zabbix.conf.php SAML settings

root-expert:
Implements SAML settings as described [here](https://www.zabbix.com/documentation/current/manual/web_interface/frontend_sections/administration/authentication#advanced_settings).

Fixes #693 
Closes #720

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -182,6 +182,10 @@ class zabbix (
   String $additional_service_params                                       = $zabbix::params::additional_service_params,
   Optional[String[1]] $zabbix_user                                        = $zabbix::params::server_zabbix_user,
   Optional[String] $zabbix_server_name                                    = $zabbix::params::zabbix_server,
+  Optional[Stdlib::Absolutepath] $saml_sp_key                             = $zabbix::params::saml_sp_key,
+  Optional[Stdlib::Absolutepath] $saml_sp_cert                            = $zabbix::params::saml_sp_cert,
+  Optional[Stdlib::Absolutepath] $saml_idp_cert                           = $zabbix::params::saml_idp_cert,
+  Hash[String[1], Variant[ScalarData, Hash]] $saml_settings               = $zabbix::params::saml_settings,
 ) inherits zabbix::params {
   class { 'zabbix::web':
     zabbix_url                               => $zabbix_url,
@@ -224,6 +228,10 @@ class zabbix (
     ldap_clientcert                          => $ldap_clientcert,
     ldap_clientkey                           => $ldap_clientkey,
     ldap_reqcert                             => $ldap_reqcert,
+    saml_sp_key                              => $saml_sp_key,
+    saml_sp_cert                             => $saml_sp_cert,
+    saml_idp_cert                            => $saml_idp_cert,
+    saml_settings                            => $saml_settings,
     manage_selinux                           => $manage_selinux,
     require                                  => Class['zabbix::server'],
   }
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -148,16 +148,28 @@
 #   Max amount of vars for GET/POST requests
 #
 # [*ldap_cacert*]
-#  Set location of ca_cert used by LDAP authentication.
+#   Set location of ca_cert used by LDAP authentication.
 #
 # [*ldap_clientcrt*]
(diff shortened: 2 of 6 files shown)


manifests/web.pp
```diff
@@ -229,6 +241,10 @@
   $ldap_clientcert                                                    = $zabbix::params::ldap_clientcert,
   $ldap_clientkey                                                     = $zabbix::params::ldap_clientkey,
   Optional[Enum['never','allow','try','demand','hard']] $ldap_reqcert = $zabbix::params::ldap_reqcert,
+  Optional[Stdlib::Absolutepath] $saml_sp_key                         = $zabbix::params::saml_sp_key,
+  Optional[Stdlib::Absolutepath] $saml_sp_cert                        = $zabbix::params::saml_sp_cert,
+  Optional[Stdlib::Absolutepath] $saml_idp_cert                       = $zabbix::params::saml_idp_cert,
+  Optional[Hash] $saml_settings                                       = $zabbix::params::saml_settings,
```
root-expert, line 247: Maybe we can enforce more strict type here? Dunno

root-expert:
Rebased with master

manifests/init.pp
```diff
@@ -71,111 +71,115 @@
...
+  $loadmodule                                          = $zabbix::params::server_loadmodule,
+  Optional[Stdlib::Absolutepath] $socketdir            = $zabbix::params::server_socketdir,
+  Boolean $manage_selinux                              = $zabbix::params::manage_selinux,
+  String $additional_service_params                    = $zabbix::params::additional_service_params,
+  Optional[String[1]] $zabbix_user                     = $zabbix::params::server_zabbix_user,
+  Optional[String] $zabbix_server_name                 = $zabbix::params::zabbix_server,
+  Optional[Stdlib::Absolutepath] $saml_sp_key          = $zabbix::params::saml_sp_key,
+  Optional[Stdlib::Absolutepath] $saml_sp_cert         = $zabbix::params::saml_sp_cert,
+  Optional[Stdlib::Absolutepath] $saml_idp_cert        = $zabbix::params::saml_idp_cert,
+  Optional[Hash[String[1], ScalarData]] $saml_settings = $zabbix::params::saml_settings,
```
bastelfreak, line 182: can you instead default to an empty hash? we try to use that as default instead of undef. we've our code guidelines at https://voxpupuli.org/docs/reviewing_pr/

root-expert: Thanks, I'll go through the guidelines!
Is `Optional[Hash]` type okay for this purpose?

bastelfreak: no, it should be `Hash[String[1], ScalarData]` and the value in params.pp should be `{}` and not undef

root-expert: Mhmm that doesn't work, because I can have a hash as a value. 

https://github.com/voxpupuli/puppet-zabbix/pull/763/files#diff-82ccb785b8e760dc28ab7067a48ebd387c93d71c401fbebb9256bfa38117014aR221-R229

bastelfreak: why doesn't that work? It's a common pattern to default to an empty hash. The datatypes with in the `Hash[]` type only apply if the hash is not empty, so that's fine in this case. Do the tests throw any error?

root-expert: Defaulting to empty hash is okay.

This is the error: 

```
Class[Zabbix::Web]: parameter 'saml_settings' entry 'security' expects a ScalarData value, got Struct
```

root-expert: `Hash[String[1], Variant[ScalarData, Hash]]` seems to be working

PR #765, opened 2021-05-05
Ensure consistent use of logtype and logfile across agent/server/proxy

shaunbugler:
LogType and LogFile were used inconsistently across agent/server/proxy, this PR ensures params, templates and tests are all consistent across the 3 services

####zabbix::agent
- logtype param was not defined in zabbix::params, also had different type enforcement that other classes
####zabbix::server
- logfile param had different type enforcement that other classes
####zabbix::proxy
- logtype had different type enforcement that other classes
####zabbix::init
- ensure we can pass in LogType to zabbix::server

Added tests to ensure we only include LogFile/LogFileSize in configuration if LogType is set to 'file'.
Also ensure we document the LogType param correctly for all the above.

Rename camp2camp as camptocamp


bastelfreak:
@shaunbugler thanks for the PR. can you please rebase against our latest master?

manifests/init.pp
```diff
@@ -109,6 +109,7 @@
   $sourceip                                 = $zabbix::params::server_sourceip,
   $logfile                                  = $zabbix::params::server_logfile,
   $logfilesize                              = $zabbix::params::server_logfilesize,
+  $logtype                                  = $zabbix::params::server_logtype,
```
bastelfreak, line 112: please also add the datatype here

root-expert (approved):

bastelfreak (approved):

bastelfreak:
@shaunbugler please take a look at the failing tests

shaunbugler:
Sorry my change to add types to init I didn't run tests before commiting, have fixed now, tests should pass

Rewerson:
Got a problem with this part:
 if $facts['systemd'] {
```
    $proxy_logtype                          = 'system'
    $proxy_logfile                          = undef
    $proxy_logfilesize                      = undef
```
Ok, puppet found systemd, but in my manifest I want to use log file. So I defined `logtype` as `file`, and needed `logfile` path (custom, not one that was defined earlier in module). But earlier also I got logfilesize, now it's broken, because there is empty value in config.
And how should I find out what parameters are now optional, and what are required?
I suggest to init $proxy_logfile and $proxy_logfilesize how they were inited before latest update.
Or if not, put this in documentation, if I want to use syslog, I don't need to touch anything. And if I want to redefine it to file, what required parameters are.

root-expert:
`logfile` doesn't have a default value according to zabbix's [documentation](https://www.zabbix.com/documentation/current/manual/appendix/config/zabbix_proxy).
So this is something you should define yourself if you set `logtype` to `file`.

As for the broken `logfilesize` feel free to provide a PR :smiley:

aredoff:
But as a result, it turns out that by setting the parameter in proxy_logtype to the file, we get a non-working agent.
Which is not obvious.

root-expert:
#772 feel free to comment @aredoff @Rewerson

PR #766, opened 2021-05-05
Support sslcertlocation & sslkeylocation for zabbix proxy

shaunbugler:
Support these params in zabbix::proxy:
- sslcalocation 
- sslcertlocation 
- sslkeylocation

zabbix::proxy has supported sslcalocation, sslcertlocation and sslkeylocation since version 2.4, these params were supported in the zabbix proxy template but never added to the zabbix::proxy class


manifests/proxy.pp
```diff
@@ -424,6 +430,8 @@
   $tmpdir                                   = $zabbix::params::proxy_tmpdir,
   $allowroot                                = $zabbix::params::proxy_allowroot,
   $include_dir                              = $zabbix::params::proxy_include,
+  $sslcertlocation_dir                      = $zabbix::params::proxy_sslcertlocation,
+  $sslkeylocation_dir                       = $zabbix::params::proxy_sslkeylocation,
```
root-expert, line 434: Can you use `Optional[Stdlib::Absolutepath]`  type here?

manifests/params.pp
```diff
@@ -386,6 +386,8 @@
   $proxy_snmptrapperfile                    = '/tmp/zabbix_traps.tmp'
   $proxy_sourceip                           = undef
   $proxy_sshkeylocation                     = undef
+  $proxy_sslcertlocation                    = '/usr/lib/zabbix/ssl/certs'
+  $proxy_sslkeylocation                     = '/usr/lib/zabbix/ssl/keys'
```
root-expert, line 390: I think zabbix doesn't have default value. That should be undef in this case.

root-expert (changes requested):
Thank you for the PR, I think `proxy_sslcalocation` is missing as well. If you could add this to this PR would be awesome!

manifests/params.pp
shaunbugler: you are correct, will update

manifests/proxy.pp
shaunbugler: Will update

templates/zabbix_proxy.conf.erb
```diff
@@ -406,20 +406,20 @@ Include=<%= @include_dir %>
 #	Location of SSL client certificates.
 #	This parameter is used only in web monitoring.
 #
-<% if @sslcertlocation %>SSLCertLocation=<%= @sslcertlocation %><% end %>
```
bastelfreak, line 409: mhm where did those variables come from? is that something that needs to be cleaned up?

shaunbugler: They were added with support for zabbix proxy 3.0, but if you check the merge while the template was updated the params were never added, this is what I am fixing

PR #770, opened 2021-05-08
Add Vault parameters

root-expert:
Adds Vault paramters to server/proxy. Specifically:

* `VaultDBPath`
* `VaultToken`
* `VaultURL`


manifests/init.pp
```diff
@@ -145,6 +145,9 @@
   $startjavapollers                                                           = $zabbix::params::server_startjavapollers,
   Integer[1, 100] $startlldprocessors                                         = $zabbix::params::server_startlldprocessors,
   $startvmwarecollectors                                                      = $zabbix::params::server_startvmwarecollectors,
+  Optional[String] $vaultdbpath                                               = $zabbix::params::server_vaultdbpath,
+  Optional[String] $vaulttoken                                                = $zabbix::params::server_vaulttoken,
+  Stdlib::HTTPUrl $vaulturl                                                   = $zabbix::params::server_vaulturl,
```
bastelfreak, line 150: what do you think about those types?
```suggestion
  Optional[String[1]] $vaultdbpath                                            = $zabbix::params::server_vaultdbpath,
  Optional[String[1]] $vaulttoken                                             = $zabbix::params::server_vaulttoken,
  Stdlib::HTTPSUrl $vaulturl                                                  = $zabbix::params::server_vaulturl,
```

root-expert: Seems fine, as longs as people don't want to put `http://` URLs into the Vault URL param

bastelfreak: I think we can prohibit access to a secret store via http. Otherwise I'm happy to argue with every user that thinks otherwise :D

root-expert: Me too! :smile:

bastelfreak (approved):

PR #772, opened 2021-05-12
Ensure correct logfile/logfilesize values when defaulting logtype to system

root-expert:
When setting `logtype` to system ensure `logfile` and `logfilesize` have correct values instead of `undef`.
Chagning logtype to file could lead to non-working agent due to these values being undef, which is not obvious.


root-expert:
Hopefully that's not backwards incompatible anymore.

LogType still defaults to system when systemd is detected, but the other params, LogFile/LogFileSize, have correct default values.

PR #776, opened 2021-07-25
Add Zabbix 5.4 support

### LINKED ISSUE #773: Zabbix SQL file location has changed
## Affected module

- Module version: 9

Use the module to install either the Zabbix server of Proxy

## What are you seeing
SQL files missing

Module to find SQL file and setup database

From Zabbix 5.4 onward, the SQL files are no longer included in the zabbix server or proxy package. They have now moved to their own package called: zabbix-sql-scripts

See: https://www.zabbix.com/documentation/current/manual/installation/upgrade_notes_540#sql_file_location
https://support.zabbix.com/browse/ZBX-19451

Thank you.

root-expert:
Adds Zabbix 5.4 support.
zabbixapi gem doesn't have support for 5.4 version yet. 
For now I passed `ignore_version` parameter. It seems to be working fine as there aren't any API changes that would break the gem. If we don't feel comfortable merging this I may open a PR on zabbixapi repo.

Closes #773 (Thanks for pointing out!)

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -6,6 +6,7 @@ Gemfile:
       - gem: 'zabbixapi'
 spec/spec_helper.rb:
   mock_with: ':mocha'
+  spec_overrides: "require 'support/acceptance/supported_versions'"
 spec/spec_helper_acceptance.rb:
   unmanaged: false
 .github/workflows/ci.yml:
--- a/lib/puppet/provider/zabbix.rb
+++ b/lib/puppet/provider/zabbix.rb
@@ -51,7 +51,8 @@ class Puppet::Provider::Zabbix < Puppet::Provider
       user: api_config['default']['zabbix_user'],
       password: api_config['default']['zabbix_pass'],
       http_user: api_config['default']['zabbix_user'],
-      http_password: api_config['default']['zabbix_pass']
+      http_password: api_config['default']['zabbix_pass'],
+      ignore_version: true
     )
     zbx
   end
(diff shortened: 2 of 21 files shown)


bastelfreak (commented):
this is backwards incompatible because the application support is being droped. Please check if we can keep it until we drop support for those old versions that still have it.

spec/supported_versions.rb
```diff
@@ -0,0 +1,3 @@
+def supported_versions
```
bastelfreak, line 1: can you move the file?
https://github.com/voxpupuli/puppet-zabbix/blob/master/spec/spec_helper_acceptance.rb#L6
we've some autoloader logic here. If you do that, you don't need to add the require to all spec files.

root-expert: First I added in `/spec/support/acceptance` but I'm using it in some unit tests as well.
Any idea how I could tackle that?

bastelfreak: indeed! https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/spec/spec_helper.rb.erb#L30
our spec_helper.rb can be extended with additional lines. You can see an example here:
* https://github.com/voxpupuli/puppet-collectd/blob/master/spec/spec_helper.rb#L20
* https://github.com/voxpupuli/puppet-collectd/commit/c5e7f6d1d00c7bce637b31118873773a4a9aa9ff#diff-b1bbc4d50c1c098ca18224cbc9519ad646dcc5e3dd912edf55610ab5bba3566eL12-L13

So I suggest:
* move the file into the acceptance dir
* update .sync.yml
* update spec_helper.rb

root-expert: Just to be sure, from .sync.yml I need to remove the `spec/spec_helper.rb` line? (along with mock_with :stuck_out_tongue: )

root-expert: My bad, i need the spec_overrides property.

bastelfreak (approved):

PR #779, opened 2021-08-05
modulesync 4.1.0

bastelfreak:
modulesync 4.1.0


root-expert:
Invalid sequence in US-ASCII again :disappointed:

root-expert (approved):

PR #782, opened 2021-08-12
Support for Zabbix Agent 2 through agent parameters

### LINKED ISSUE #692: Add Support for Zabbix Agent 2
The current version of this module does not support Zabbix Agent 2.
Agent 2 has much more functionality and we would like to switch to it. 
Please add new agent support to your wonderful module.

sgnl05:
This PR adds support for using Zabbix Agent 2, without adding much complexity to the module.

Users wanting to use Zabbix Agent 2 will have to set the following parameters in Hiera:

```yaml
zabbix::agent::agent_configfile_path: '/etc/zabbix/zabbix_agent2.conf'
zabbix::agent::include_dir: '/etc/zabbix/zabbix_agent2.d'
zabbix::agent::pidfile: '/var/run/zabbix/zabbix_agentd2.pid'
zabbix::agent::servicename: 'zabbix-agent2'
zabbix::agent::zabbix_package_agent: 'zabbix-agent2'
```

Fixes #692


kenyon:
Could you use EPP instead of ERB for the new templates? We prefer to use EPP.

sgnl05:
Sure! I'll start working on them tomorrow.

root-expert:
Hello, thanks for your contribution!

Can you add a section in the readme on how to configure agent2 both by class parameters and with hiera?

Also we have to consider adding acceptance tests for agent v2.

manifests/agent.pp
```diff
@@ -291,23 +291,40 @@
...
   if $manage_startup_script {
-    zabbix::startup { $servicename:
-      pidfile                   => $pidfile,
-      agent_configfile_path     => $agent_configfile_path,
-      zabbix_user               => $zabbix_user,
-      additional_service_params => $additional_service_params,
-      service_type              => $service_type,
-      service_name              => 'zabbix-agent',
-      require                   => Package[$zabbix_package_agent],
+    if $agent_configfile_path =~ /agent2/ {
```
bastelfreak, line 294: please set a variable within this if block for the service_name and don't duplicate the zabbix::startup block.

templates/zabbix-agent2-systemd.init.erb
```diff
@@ -0,0 +1,16 @@
+# THIS FILE IS MANAGED BY PUPPET
```
bastelfreak, line 1: for the new agent, do we need to duplicate the systemd unit file? isn't there one in the package we can use?

manifests/agent.pp
sgnl05: I initially used $servicename for the service_name parameter, but $servicename varies for different systems (which is set in params.pp), and the service_name variable is used for getting the template file name. I'll probably have to modify startup.pp to fix it.

templates/zabbix-agent2-systemd.init.erb
sgnl05: Do you mean that we should consider not managing the systemd unit file? If yes, then I agree. :)

bastelfreak: yes. if there is one we should use it. If it has any flaws we can use a drop_in to overwrite the specific settings.

manifests/agent.pp
sgnl05: Nvm, I get it now. I'll fix it shortly

manifests/agent.pp
```diff
@@ -289,25 +289,34 @@
     }
   }
 
-  # Ensure that the correct config file is used.
+  # Ensure that the correct startup file is used.
   if $manage_startup_script {
+    $service_name = $servicename ? {
```
bastelfreak, line 294: selector statements <3

templates/zabbix-agent2-aix.init.epp
```diff
@@ -0,0 +1,46 @@
+<%- | String $pidfile,
```
bastelfreak, line 1: same as with the init file, do we really have to ship this one via puppet?

templates/zabbix-agent2-debian.init.erb
```diff
@@ -0,0 +1,68 @@
+#! /bin/sh
```
bastelfreak, line 1: are there any supported distributions that still use this legacy style and not the systemd unit?

templates/zabbix-agent2-redhat.init.erb
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
```
bastelfreak, line 1: same here, do we really need this file anymore?

templates/zabbix-agent2-debian.init.erb
sgnl05: Probably not! Should I remove these files and relevant code?

bastelfreak: yes please. let's keep the complexity to the minimal required one.

manifests/agent.pp
```diff
@@ -325,6 +339,18 @@
...
+  # Otherwise use the defined erb/epp template, or use default
+  if $agent_configfile_epp and $agent_configfile_template {
+    fail('Cannot supply both agent_configfile_epp and agent_configfile_template templates for config file.')
+  } elsif $agent_configfile_template {
+    $config_content = template($agent_configfile_template)
+  } elsif $agent_configfile_epp {
+    $config_content = epp($agent_configfile_epp)
+  } else {
+    $config_content = template('zabbix/zabbix_agentd.conf.erb')
+  }
```
root-expert, line 352: I'm not sure I understand the logic, but I don't think the user should define where to pick the template (either epp or erb). That should be predefined in the module.

sgnl05: I used the same logic here as puppetlabs-ntp uses. I can however change it if it's not the preferred way.

templates/zabbix-agent2-debian.init.erb
sgnl05: I've cleaned up the files and set the manage_startup_script variable to default to false in params.pp. However, I see some of the tests are failing now because systemd is the only supported startup type. I've looked at fixing the tests for this, but I fear I'm way over my head. 🙄

manifests/agent.pp
sgnl05: Selector to the rescue

sgnl05:
I think I've done what I can now. Some tests are failing due to cleanup of init script for unsupported operating system versions. We should probably just remove those tests, but I'm always struggling with tests and I guess I'll just have to ask for help here.

templates/zabbix-agent2-systemd.init.erb
sgnl05: This should be fixed. Please let me know if my solution is good enough :)

manifests/agent.pp
root-expert: Isn't that kind of breaking change? 
Users were able to override service name before, now we default to either `zabbix-agent` or `zabbix-agent2`

sgnl05: Users select the servicename by setting $servicename and there's no change there.
The variable $service_name is only used when calling startup.pp. The parameter sent to startup.pp was earlier hardcoded to 'zabbix-agent'.

root-expert: oops you're right

root-expert:
My only concern is that a global switch may be better for the end user. 

Maybe a boolean paramater `use_agent_v2` will automatically set the correct default values (e.g package name, service name) and install v2 package.
That may increase the module complexity though :(

Another thing is that, if a user has zabbix agent v1 already installed and wants to switch to v2, the module should take care of purging v1 package (and vice-versa).

Thoughs?

sgnl05:
I think in general that a module that installs a package, sets a config file and manages the service should have parameters that are configurable so that alternative packages/config templates/service names can be used. It would give the module the flexibility needed to use Zabbix Agent 2, without increasing complexity. 

This PR became way to complex, and I'm unable to fix the tests that fail because of it.

I'd like to close the PR, and submit a new one that only focuses on the package name, config template and the service name,

PR #785, opened 2021-08-24
Added details to host network interface to configure SNMP hosts

oraziobattaglia:
#### Pull Request (PR) description

Following the pull request #697 we add an hash of interface options to configure some details about SNMP.
We want to use to add SNMP host with community and other details.

Example

```
  zabbix_host { 'SNMPServer':
    ipaddress     => '192.168.1.1',
    use_ip        => true,
    port          => 161,
    groups        => ['SNMP servers'],
    group_create  => true,
    templates     => ['Template Module ICMP Ping'],
    macros        => [],
    proxy         => '',
    interfacetype => 2,
    interfacedetails => {"version" => 2, "bulk" => 0, "community" => "public"},
  }
```

Using interfacetype = 2 without details we have the error

```
{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid params.","data":"Incorrect arguments passed to function."},"id":42907}
```

#### This Pull Request (PR) fixes the following issues

No issues opened, follow pull request #697

PR #793, opened 2021-09-14
Add code to ensure interfacedetails is always a hash

oraziobattaglia:

This PR follow PR #785 to ensure the host object idempotency. When interface details are empty the zabbix API returns an empty array and I need to convert it in an empty hash.

#### Example

This is returned by the api call:

```json
      "interfaces": [
        {
          "interfaceid": "36",
          "type": "1",
          "main": "1",
          "ip": "10.10.10.1",
          "port": "10050",
          "useip": "1",
          "details": []
        }
      ]
```

and the error on puppet is:

```
interfacedetails changed [] to {} (corrective)
```

### DIFF
--- a/lib/puppet/type/zabbix_host.rb
+++ b/lib/puppet/type/zabbix_host.rb
@@ -54,6 +54,7 @@ Puppet::Type.newtype(:zabbix_host) do
 
   newproperty(:interfacedetails) do
     desc 'Additional interface details.'
+
     def insync?(is)
       is.to_s == should.to_s
     end
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -152,7 +152,7 @@ class zabbix::agent (
   $zbx_templates                                       = $zabbix::params::agent_zbx_templates,
   Array[Hash] $zbx_macros                              = [],
   Integer[1,4] $zbx_interface_type                     = 1,
-  Hash[String, Any] $zbx_interface_details             = {},
+  Variant[Array, Hash] $zbx_interface_details          = [],
   $agent_configfile_path                               = $zabbix::params::agent_configfile_path,
   $pidfile                                             = $zabbix::params::agent_pidfile,
   $servicename                                         = $zabbix::params::agent_servicename,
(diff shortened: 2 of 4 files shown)


bastelfreak:
hey @oraziobattaglia, thanks for the PR. do you think there is an easy way to add tests for this?

root-expert:
Hey @oraziobattaglia according to the API [reference](https://www.zabbix.com/documentation/current/manual/api/reference/hostinterface/object) details is supposed to be an Array. Maybe it returns an array of hashes every time?

oraziobattaglia:
Hi @root-expert me and my colleague working on this have seen that if interfacedetails has data the API returns an hash of data like

```
          "details": {
            "version": "2",
            "bulk": "1",
            "community": "{$SNMP_COMMUNITY}"
          }
```

In case interfacedetails has no data the API returns an empty array like

```
          "details": []
```

I don't know why, maybe is a question related to the API, but the resource isn't idempotent.

So the solution I found here is to declare a variant type for interfacedetails:

```
Variant[Array, Hash] $interfacedetails = [],
```

We also tried to declare interfacedetails as an array of hashes and seems to work fine too:

```
Array[Hash] $interfacedetails = [],
```

But in this second case I don't full understand how it works!! :-O

Let me know if you have suggestions.
Thanks for the help.

root-expert:
I would say that this could be a Zabbix bug or incorrect documentation. Can you try raising an issue to their [Jira](https://support.zabbix.com/) before continuing further with this PR?

The Variant solution seems okay to me. The `Array[Hash]` we have to ensure that passing
```
"details": [
  { "foo": "bar" }
]
```

to the API is okay, or the API throws an error

oraziobattaglia:
Hi @root-expert seems it's a known bug, here they have similar problems with ansible https://github.com/ansible-collections/community.zabbix/pull/51

oraziobattaglia:
The interface details as array of hashes doesn't work:

```
"details": [ { "foo": "bar" } ]
```

Puppet doesn't converge and there is always the error:

```
expected "details": [ { "foo": "bar" } ]
received "details": { "foo": "bar" }
```

This is because, I think, there is a bug on the zabbix API. When interface details are given as hash the API returns an hash of data (not an array of hash). When interface details are not given the API returns an empty array [].

I add a test, test4.example.com, that show that an empty interface details returns an empty array.

If you have suggestions they are welcome!
Thank you

root-expert (approved):

PR #797, opened 2021-09-24
Zabbix 42 and 44 also work fine

Fabian1976:
Zabbix 4.2 and 4.4 work fine with this code. Tested

It actually works a lot better with these version then the previous major release (8) because of the updated dependency of the zabbix-api gem.

### DIFF
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -192,7 +192,7 @@ class zabbix::web (
   if $manage_resources {
     # Determine correct zabbixapi version.
     case $zabbix_version {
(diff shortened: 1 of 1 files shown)


root-expert:
Hello @Fabian1976 , 4.2 and 4.4 are EoL and we no longer accept PRs and issues for those versions.
The support for these versions was dropped at #750

Fabian1976:
So why is 5.2 still in the code? That is also no longer supported by Zabbix.

And why so strict with the code?

PR #801, opened 2021-10-12
Respect manage_repo for RedHat family

fraenki:
In https://github.com/voxpupuli/puppet-zabbix/pull/736 the CentOS SCL repo was added by installing `centos-release-scl`. However, this breaks repository management for folks that have `$manage_repo` set to `false`.

This PR ensure that the SCL repo package is only installed when `$manage_repo` is set to `true`.

Another approach would be to move the installation of the `centos-release-scl` package from zabbix::web to zabbix::repo. Let me know what you think.

Having two modules trying to manage the SCL repo leads to all kinds of issues, i.e. duplicate declarations.


feltra:
Currently this is adding centos-release-scl when it already have repo.zabbix.com. When trying to install the agent, this repo doesn't seems to be required w since you already have zabbix repo installed.

Issue #812, opened 2022-01-25
DenyKey active by default even if denykey => undef,


tunisiano187:
<!--
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 7.5
- Ruby: ?
- Distribution: Ubuntu
- Module version: latest

## How to reproduce (e.g Puppet code you use)
class { 'zabbix::agent':
     server => 'xxxx',
     denykey => undef,
}

## What are you seeing
DenyKey not recognised at start of the zabbix-agent

## What behaviour did you expect instead
when using denykey => undef, don't have it in the config file

PR #813, opened 2022-01-29
Default denykey to undef

root-expert:
Having a default value makes it impossible to undef the parameter, since undefing leads to the default value being used.
As per [zabbix's](https://www.zabbix.com/documentation/5.0/en/manual/appendix/config/zabbix_agentd) documentation this parameter is no mandatory, so undef as default value should be okay.

#812

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -297,7 +297,7 @@ class zabbix::params {
   $agent_buffersize                         = '100'
   $agent_debuglevel                         = '3'
   $agent_allowkey                           = undef
-  $agent_denykey                            = 'system.run[*]'
+  $agent_denykey                            = undef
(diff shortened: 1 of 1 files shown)


bastelfreak:
This fails because of missing quotes here?
https://github.com/voxpupuli/puppet-zabbix/blame/37f1a58645d97012ee7990becb600ec98bb2dc66/manifests/repo.pp#L121

root-expert:
It fails on `$facts['os']` but it seems that it's not specific as it fails on modules [outside](https://github.com/voxpupuli/puppet-zabbix/runs/4996566163?check_suite_focus=true#step:4:1599) of it?

bastelfreak:

PR #815, opened 2022-02-23
Zabbix-Agent2 basic support (#1)

dsoltero:
* allow configuration file that is compatible with zabbix_agent2 (remove some of the hardcoded values)
* add additional value validation
* setup Integer values as integers and not strings


templates/zabbix_agentd.conf.erb
```diff
@@ -216,38 +205,32 @@ MaxLinesPerSecond=<%= @maxlinespersecond %>
...
 ### Option: AllowRoot
 #	Allow the agent to run as 'root'. If disabled and the agent is started by 'root', the agent
 #	will try to switch to the user specified by the User configuration option instead.
 #	Has no effect if started under a regular user.
 #	0 - do not allow
 #	1 - allow
 #
-AllowRoot=<%= @allowroot %>
-<% end %>
+<% if @allowroot && @kernel != 'windows' %>AllowRoot=<%= @allowroot %><% end %>
```
root-expert, line 220: The if here was supposed to be on top to remove the comments as well btw

Same for the rest of these

root-expert (commented):
Thanks for the PR!

Generally this PR is far from what the title suggest IMHO.
Especially the changes on `agent.pp` although they are welcomed they should probably be moved to another PR.

templates/zabbix_agentd.conf.erb
```diff
@@ -69,27 +64,21 @@ DebugLevel=<%= @debuglevel %>
 #       If no AllowKey or DenyKey rules defined, all keys are allowed.
 #       Unless another system.run[*] rule is specified DenyKey=system.run[*] is added by default.
 #
-# Mandatory: no
-# Default:
-<% if @denykey %>DenyKey=<%= @denykey -%><% end %>
-<% end %>
+<% if @zabbix_version.to_f >= 5.0 && @denykey %>DenyKey=<%= @denykey -%><% end %>
 
-<% if @zabbix_version.to_f < 5.0 %>
```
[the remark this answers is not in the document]
dsoltero: That configuration was existing, I am just moving the "block" of the template file from: 
```
<% if @zabbix_version.to_f < 5.0 %>
# Option 
Option= <%= $option %>
<% end %>
```
to: 
```
# Option 
<% if @zabbix_version.to_f < 5.0 %>Option= <%= $option %><% end %>
```

templates/zabbix_agentd.conf.erb
dsoltero: Right, but do not understand the issue of having more comments.

dsoltero:
> Thanks for the PR!
> 
> Generally this PR is far from what the title suggest IMHO. Especially the changes on `agent.pp` although they are welcomed they should probably be moved to another PR.

Sure, I can split this into a number of smaller changes if that helps. That should allow the review of the code to be simpler.

PR #816, opened 2022-02-23
Fix typo in repo.pp

root-expert:


root-expert:
The tests won't pass either way but still that needs fixing

bastelfreak (approved):

PR #818, opened 2022-03-09
 Fix #817 - startup_script fails for windows

### LINKED ISSUE #817: startup_script fails for windows
This bug is caused due to a case sensitivity issue with the service name for windows.  On windows the service name is capitalized while everything else it is lowercase.   This causes the service to not start because the fail function is executed.  A case insensitive comparison will fix this issue. 

https://github.com/voxpupuli/puppet-zabbix/blob/master/manifests/startup.pp#L32

https://github.com/voxpupuli/puppet-zabbix/blob/master/manifests/params.pp#L103

## What are you seeing
The catalog fails: `we currently only support a title that contains agent or server`

The catalog to complete sucessfully

logicminds:
* Previously the windows agent puppet code did not handle windows
     properly and failed the catalog build due to a case sensitive issue.
     This fixes that by downcase the agent name first.


bastelfreak:
thanks for the PR. on which windows versions did you test / are they missing in https://github.com/voxpupuli/puppet-zabbix/blob/master/metadata.json#L143-L149 ?

logicminds:
Yea this was on 2019, but it is a code issue and not a platform one.  Anyways 2019 work great so I think adding 2019 is applicable.  Do you want me to update the metadata to include that?

bastelfreak:
yes, please add it :) (and if you've really much spare time and could submit windows facts to https://github.com/voxpupuli/facterdb/pulls I will add you to my chocolate-giftlist-for-future-conferences)

bastelfreak:

PR #819, opened 2022-03-10
Ubuntu: Install lsb-release during CI

bastelfreak:

### DIFF
--- a/spec/setup_acceptance_node.pp
+++ b/spec/setup_acceptance_node.pp
@@ -1,0 +1,6 @@
+# Needed for os.distro.codebase fact on ubuntu 16/18 on puppet 6
+if $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['puppetversion'], '7.0.0') < 0 {
+  package{'lsb-release':
+    ensure => present,
+  }
+}


spec/spec/setup_acceptance_node.pp
```diff
@@ -0,0 +1,6 @@
+# Needed for os.distro.codebase fact on ubuntu 16/18 on puppet 6
+if $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['puppetversion'], '7') <= 0 {
```
smortex, line 2:
```suggestion
if $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['puppetversion'], '7.0.0') < 0 {
```

I remember some weirdness comparing `x` and `x.0.0` (probably fixed now) so being explicit I am sure the behavior is the good one and with just `<` I know we won't install in on Puppet 7.

smortex (approved):
LGTM!  I guess someone will have to contribute Archlinux support in the apache module :wink: :grin: :+1:

bastelfreak:
yeah, will do that later :)

PR #828, opened 2022-07-28
Add sensitive support

### LINKED ISSUE #440: Manage resources leaks API password in resource types
Currently the way `zabbix_host` is configured, it will inject `$zabbix_api_pass` into the resource with the property name `zabbix_pass`. This results in a password leak when puppet reports / resources can be inspected by third parties. At our setup, we run puppetboard without authentication. Other packages don't put clear text passwords in reports and resources, so this setup mostly works for us.

- Puppet: 4.10.4
- Ruby: ?
- Distribution: Ubuntu 16.04
- Module version: 4.1.3

## What are you seeing
When running the following PQL, the result includes the API credentials:
```
resources { type = "Zabbix_host" }
```

No passwords being logged.

Other modules (e.g. puppetlabs/mongodb) write the credentials to a file (`~/.mongorc.js`). This way, there's no need to communicate the credentials through the resources.

bdeferme:
Adds sensitive support for passwords

Fixes #440


manifests/database/postgresql.pp
```diff
@@ -65,9 +71,9 @@
   }
 
   exec { 'update_pgpass':
-    command => "echo ${database_host}:5432:${database_name}:${database_user}:${database_password} >> /root/.pgpass",
+    command => "echo ${database_host}:5432:${database_name}:${database_user}:${database_password_unsensitive} >> /root/.pgpass",
```
alexjfisher, line 74: Are you not still exposing the password when Puppet is run with `--debug`?

```suggestion
    command => Sensitive("echo ${database_host}:5432:${database_name}:${database_user}:${database_password_unsensitive} >> /root/.pgpass"),
```

(But I don't think I would bother making this conditional on whether `$database_password` is `Sensitive`.)

You also need to do the same with `unless`.

alexjfisher: (I also wonder why `file_line` hasn't been used here)

bdeferme: Correct, didn't think of that. I updated the exec with `Sensitive` for the `command` + `unless`. Maybe this part stems from before `file_line` was even available?

bdeferme:
Rebased on newer master

teluq-pbrideau:
I would also be interrested in this feature, can I help in any way?
@bdeferme Could you rebase from master to fix the test about `web.pp` and make it go forward?

bdeferme:
I rebased but some tests (some related to my code, some not) are failing. I  can't find the time atm too look at this. If you could @teluq-pbrideau go ahead. Otherwise I will maybe find some time next year :-)

teluq-pbrideau:
Thanks @bdeferme, as you see, i’ve duplicated to a new PR, where I will be able to work easily.
I’m not sure it is the best way to collaborate, feel free if you think I should do otherwise

PR #831, opened 2022-08-03
Escape string inside update_pgpass command

### LINKED ISSUE #830: Exec[update_pgpass] failed

- Puppet: 6.27.1-1focal
- Ruby: 2.5.9p229
- Distribution: Ubuntu 20.04
- Module version: master (currently v9.2.0)

Use Postgresql and set a custom password with special characters such as `&`.

```
class { 'apache':
    mpm_module => 'prefork',
  }
  class { 'apache::mod::php': }
  class { 'postgresql::server': }
  class { 'zabbix':
    zabbix_url        => 'zabbix.example.com',
    database_password => 'secret&password',
  }
  ```

## What are you seeing

```
Error: sh: 1: password: not found
localhost:5432:zabbix_server:zabbix_server:secret

Error: /Stage[main]/Zabbix::Database::Postgresql/Exec[update_pgpass]/returns: change from 'notrun' to ['0'] failed: sh: 1: password: not found
localhost:5432:zabbix_server:zabbix_server:secret
```

```
Notice: /Stage[main]/Zabbix::Database::Postgresql/Exec[update_pgpass]/returns: executed successfully
```

The `Exec[update_pgpass]` is defined here: https://github.com/voxpupuli/puppet-zabbix/blob/master/manifests/database/postgresql.pp#L67

``` command => "echo ${database_host}:5432:${database_name}:${database_user}:${database_password} >> /root/.pgpass",```

As we can see, the echoed string is not protected/escaped, resulting in some characters breaking the command.

Workaround:

```command => "echo \"${database_host}:5432:${database_name}:${database_user}:${database_password}\" >> /root/.pgpass",```

It is interesting to note that the `unless =>` argument was already protected/escaped.

### PR DESCRIPTION
**Pull request by src386**

The `update_pgpass` resource may fail when using strong `${database_password}` that contain special characters such as `&` because the echoed string is not escaped. This Pull request adds ~~double quotes~~ `shell_escape()` (as suggested by @smortex).

- Fixes #830

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -65,9 +65,9 @@ class zabbix::database::postgresql (
   }
 
   exec { 'update_pgpass':
-    command => "echo ${database_host}:5432:${database_name}:${database_user}:${database_password} >> /root/.pgpass",
+    command => "echo ${database_host}:5432:${database_name}:${database_user}:${shell_escape($database_password)} >> /root/.pgpass",
     path    => "/bin:/usr/bin:/usr/local/sbin:/usr/local/bin:${database_path}",
-    unless  => "grep \"${database_host}:5432:${database_name}:${database_user}:${database_password}\" /root/.pgpass",
+    unless  => "grep ${database_host}:5432:${database_name}:${database_user}:${shell_escape($database_password)} /root/.pgpass",
     require => File['/root/.pgpass'],
   }
 

### CODE REVIEW

File: manifests/database/postgresql.pp [hunk 1]
```diff
@@ -65,7 +65,7 @@
   }
 
   exec { 'update_pgpass':
-    command => "echo ${database_host}:5432:${database_name}:${database_user}:${database_password} >> /root/.pgpass",
+    command => "echo \"${database_host}:5432:${database_name}:${database_user}:${database_password}\" >> /root/.pgpass",
```
smortex, line 68: This will shoke if the password has `"` char, right?  `shell_escape()` from stdlib should be safer in this case.

src386: Thanks, I was not aware of this function. I just updated the Pull Request.

manifests/database/postgresql.pp
```diff
@@ -65,9 +65,9 @@
   }
 
   exec { 'update_pgpass':
-    command => "echo ${database_host}:5432:${database_name}:${database_user}:${database_password} >> /root/.pgpass",
+    command => "echo ${database_host}:5432:${database_name}:${database_user}:${shell_escape($database_password)} >> /root/.pgpass",
     path    => "/bin:/usr/bin:/usr/local/sbin:/usr/local/bin:${database_path}",
-    unless  => "grep \"${database_host}:5432:${database_name}:${database_user}:${database_password}\" /root/.pgpass",
+    unless  => "grep ${database_host}:5432:${database_name}:${database_user}:${shell_escape($database_password)} /root/.pgpass",
```
smortex, line 70: I can be even more creative :smiling_imp: :

```sh-session
romain@desktop-fln40kq ~ % sudo -u postgres createdb 'foo bar'
romain@desktop-fln40kq ~ % sudo -u postgres createdb 'foo:bar'
romain@desktop-fln40kq ~ % sudo -u postgres createuser 'foo bar baz'
romain@desktop-fln40kq ~ % sudo -u postgres createuser 'foo:bar:baz'
```

The spaces will break the command, and according to https://www.postgresql.org/docs/current/libpq-pgpass.html:

> If an entry needs to contain `:` or `\`, escape this character with `\`.

So I think we should also escape the `database_name`, `database_user` and `database_password` accordingly (I do not think these chars are valid for a hostname), and then shell-escape the whole pattern:

```puppet
$escaped_database_name     = $database_name.regsubst('[\\\\:]', '\\\\\\0', 'G')
$escaped_database_user     = $database_user.regsubst('[\\\\:]', '\\\\\\0', 'G')
$escaped_database_password = $database_password.regsubst('[\\\\:]', '\\\\\\0', 'G')
```
```suggestion
    unless  => "grep ${"${database_host}:5432:${escaped_database_name}:${escaped_database_user}:${escaped_database_password}".shell_escape} /root/.pgpass",
```

I only tested the pattern substitution (which required much escaping than i though), so we probably need to test this for real with users with annoying chars in their name to be sure it is fine.

But all this is probably overkill, maybe we can just detect "forbidden" chars and raise an error because nobody wants to do this kind of things?

xchotard-talentsoft: In my opinion postgresql admins should know that they must avoid special characters in databases names and usernames. This is a bit different for passwords than can be sometimes randomly generated or have strong complexity requirements. In my case I used `pwgen` to get a strong password.

They should also know that escape characters in passwords will break everything and must be avoided. So I think this is reasonable to keep the fix as simple as possible.

ekohl:
In https://github.com/voxpupuli/puppet-zabbix/pull/904 I've taken another approach where I provide all the data via env vars, removing the need for the pgpass file altogether.

PR #833, opened 2022-08-04
Drop EoL Ubuntu 16.04

root-expert:
Ubuntu 16.04 has been EoL since 30/04/21

### DIFF
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -151,7 +151,7 @@ class zabbix::web (
   # zabbix frontend 5.x is not supported, among others, on stretch and xenial.
   # https://www.zabbix.com/documentation/current/manual/installation/frontend/frontend_on_debian
   if $facts['os']['name'] in ['ubuntu', 'debian'] and versioncmp($zabbix_version, '5') >= 0 {
-    if versioncmp($facts['os']['release']['major'], '16.04') == 0 or versioncmp($facts['os']['release']['major'], '9') == 0 {
+    if versioncmp($facts['os']['release']['major'], '9') == 0 {
       fail("${facts['os']['family']} ${$facts['os']['release']['major']} is not supported for zabbix::web")
     }
   }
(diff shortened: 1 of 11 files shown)


manifests/web.pp
```diff
@@ -151,7 +151,7 @@
   # zabbix frontend 5.x is not supported, among others, on stretch and xenial.
   # https://www.zabbix.com/documentation/current/manual/installation/frontend/frontend_on_debian
   if $facts['os']['name'] in ['ubuntu', 'debian'] and versioncmp($zabbix_version, '5') >= 0 {
-    if versioncmp($facts['os']['release']['major'], '16.04') == 0 or versioncmp($facts['os']['release']['major'], '9') == 0 {
+    if versioncmp($facts['os']['release']['major'], '9') == 0 {
```
ekohl, line 154: Should it also drop Debian 9, which we also consider unsupported in Vox Pupuli modules.

manifests/web.pp
```diff
@@ -231,7 +231,7 @@
       # Check OS release for proper prefix
       case $facts['os']['name'] {
         'Ubuntu': {
-          if versioncmp($facts['os']['release']['major'], '16.04') >= 0 {
+          if versioncmp($facts['os']['release']['major'], '18.04') >= 0 {
```
ekohl, line 234: I think this whole condition can be dropped. You should never get to the `else` branch. Similar for Debian below. The default branch can't be reached so the whole case can be replaced with `$php_db_package = "php-${db}"`.

ekohl (changes requested):

manifests/web.pp
bastelfreak: cleaned it up

manifests/web.pp
bastelfreak: I did like to drop it on another PR

PR #834, opened 2022-08-09
Modulesync 5.3.0

ekohl:


ekohl:
I'm still bothered by this bit:
https://github.com/voxpupuli/puppet-zabbix/blob/f4796f1439453610bc4f483a94247f6530313551/spec/defines/startup_spec.rb#L14-L17
The loop there doesn't work as you'd expect. You should really create a `context`, otherwise you only end up testing the last value I think. However, I don't have time now to dive into this.

ekohl:
Archlinux fails because there's no Apache version set. I'm not sure how that ever worked.

ekohl:
One problem is that this changes Gentoo to have openrc as the init system (which is the default), but the tests assume systemd. I'm not sure how to deal with that.

Rathios:
Threw in some additions here that fix the remainder of the tests for me: https://github.com/ekohl/puppet-zabbix/pull/1

bastelfreak (approved):

ekohl:
Debian just fails on this when using Puppet 6:
```
         	Error: Could not update: Execution of '/opt/puppetlabs/puppet/bin/gem install -v 5.0.0-alpha1 --no-document zabbixapi' returned 1: ERROR:  Error installing zabbixapi:
         		The last version of public_suffix (< 6.0, >= 2.0.2) to support your Ruby & RubyGems was 4.0.7. Try installing it with `gem install public_suffix -v 4.0.7` and then running the current command again
         		public_suffix requires Ruby version >= 2.6. The current ruby version is 2.5.0.
```

I recall that old Rubygems versions had issues with downgrading. Not sure what the best course of action is here, but that's just broken.

bastelfreak:
:+1: for dropping Puppet 6. It's EoL in a few months anyways.

PR #835, opened 2022-08-09
puppetlabs/apache: Require 8.x; switch apache ports from string to int

bdeferme:

Stop breaking the upstream apache rspecs.

Recently we started seeing:

`Evaluation Error: Error while evaluating a Resource Statement, Apache::Vhost[xxx_nonssl]: parameter 'port' expects a value of type Undef, Array[Stdlib::Port = Integer[0, 65535]], or Stdlib::Port = Integer[0, 65535], got String (file: /repo/spec/fixtures/modules/zabbix/manifests/web.pp, line: 416) on node 022551739aa6.xxx`

This should fix that.


smortex (approved):
Wow, this module is quite broken at least on the CI side :scream: 

That PR looks right as far as I am concerned.

metadata.json
```diff
@@ -18,7 +18,7 @@
     },
     {
       "name": "puppetlabs/apache",
-      "version_requirement": ">= 1.6.0 < 7.0.0"
+      "version_requirement": ">= 8.0.0 < 9.0.0"
```
ekohl, line 21: It wasn't needed to change the lower bound so strict. This can cause problems for users who haven't updated all their yet. The old versions accepted integers just fine.

bdeferme: I must agree that the lower bound is quite aggressive 😄

PR #836, opened 2022-08-10
Disable server acceptance tests on Arch Linux

bastelfreak:
Arch Linux is currently only supported for the zabbix-agent.


bastelfreak:
We drop Ubuntu 16.04 in https://github.com/voxpupuli/puppet-zabbix/pull/833. We can ignore the failed Ubuntu 16.04 jobs in this PR.

sebastianrakel (approved):

PR #840, opened 2022-09-02
Add repo for Ubuntu on aarch64

emetriqChris:
Zabbix now provides Ubuntu Focal deb packages for the arm64 architecture:
https://support.zabbix.com/browse/ZBXNEXT-5982

They can be found under a different URL than "usual", for example:
https://repo.zabbix.com/zabbix/5.2/ubuntu-arm64/

so the repo.pp manifest has to be adapted to refelct this.

Be careful: Right now, Zabbix only provides aarch64 packages for version 5.0 and higher.

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -120,6 +120,11 @@ class zabbix::repo (
         } else {
           if ($facts['os']['distro']['id'] == 'Raspbian') {
             $operatingsystem = 'raspbian'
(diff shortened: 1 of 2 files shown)


ekohl (changes requested):
The tests are failing. Would you mind fixing them?

Looks like since Zabbix 5.5 there's also `debian-arm64`.

evgeni:
#908 was merged and contains this.

PR #843, opened 2022-09-26
On RHEL systems do not try to install the centos-release-scl package as its not present and fails

### LINKED ISSUE #823: The centos-release-scl package does not exist for RHEL

- Puppet: 7.17.0
- Ruby: 2.7.6
- Distribution: RHEL 7
- Module version: 9.2.0

```
class{ 'zabbix::agent':
    zabbix_package_state => present,
    zabbix_version       => '5.4',
    manage_firewall      => false,
    manage_repo          => true,
    service_ensure       => running,
    service_enable       => true,
    server               => $server,
    hostmetadata         => '',
    hostname             => $facts['networking']['fqdn'],
    serveractive         => "${server}:10051",
  }
```

## What are you seeing

The `centos-release-scl` package is attempted to be installed, which fails because it does not exist on RHEL 7.

The `centos-release-scl` package is skipped for RHEL 7.

## Output log

```
Could not update: Execution of '/usr/bin/yum -d 0 -e 0 -y install centos-release-scl' returned 1: Error: Nothing to do
/Stage[main]/Zabbix::Repo/Package[zabbix-required-scl-repo]/ensure | change from 'purged' to 'latest' failed: Could not update: Execution of '/usr/bin/yum -d 0 -e 0 -y install centos-release-scl' returned 1: Error: Nothing to do
```

Note that this only affects RHEL 7 specifically, the 9.2.0 release works fine on RHEL 8. 

It looks like this bug has been around for a while, but it was previously isolated only to the `zabbix::web` class, so just using `zabbix::agent` sidestepped the issue. Since #801 the code was moved to the `zabbix::repo` class, which affects everything using the managed repo on RHEL 7. 

It looks like this conditional just needs to check for the OS name to omit RHEL: https://github.com/voxpupuli/puppet-zabbix/blob/f4796f1439453610bc4f483a94247f6530313551/manifests/repo.pp#L85

anthonysomerset:

On RHEL systems do not try to install the centos-release-scl package as its not present and fails it

I chose to specifically chose to exclude Red Hat rather than to only trigger when CentOS is detected, i am unclear if the scl release will install on other variants like Rocky or AlmaLinux or not - and figured it was safer to only deal with the specific use case - otherwise this could be changed if required

Fixes #823

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -81,8 +81,8 @@ class zabbix::repo (
             priority => '1',
           }
         }
-
-        if ($facts['os']['release']['major'] == '7' and versioncmp($zabbix_version, '5.0') >= 0) {
+        
+        if ($facts['os']['name'] != 'RedHat' $facts['os']['release']['major'] == '7' and versioncmp($zabbix_version, '5.0') >= 0) {
           package { 'zabbix-required-scl-repo':
             ensure => 'latest',
             name   => 'centos-release-scl',
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -82,7 +82,7 @@ class zabbix::repo (
           }
         }
         
-        if ($facts['os']['name'] != 'RedHat' $facts['os']['release']['major'] == '7' and versioncmp($zabbix_version, '5.0') >= 0) {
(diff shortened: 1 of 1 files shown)


feltra:
Any update on this ?

anthonysomerset:
the failing check appears to be on unrelated code?

evgeni:
this has been fixed as part of #887

PR #848, opened 2022-11-14
Certificate authentication in zabbix_host resource

teluq-pbrideau:
inspired from #700

I need to authenticate with certificate in my infrastructure, this PR allow to push the configuration to the zabbix server via «storeconfigs»

As stated in #700, there is an issue with setting the PSK through the API as PSK values are write only since zabbix 5.4.  So I didn’t bother to dig deeper on this part, as I don’t require it in my infrastructure.

### DIFF
--- a/lib/puppet/provider/zabbix_host/ruby.rb
+++ b/lib/puppet/provider/zabbix_host/ruby.rb
@@ -13,7 +13,7 @@ Puppet::Type.type(:zabbix_host).provide(:ruby, parent: Puppet::Provider::Zabbix)
         selectInterfaces: %w[interfaceid type main ip port useip details],
         selectGroups: ['name'],
         selectMacros: %w[macro value],
-        output: %w[host proxy_hostid]
+        output: %w[host proxy_hostid tls_accept tls_connect tls_issuer tls_subject]
       }
     )
 
@@ -38,7 +38,11 @@ Puppet::Type.type(:zabbix_host).provide(:ruby, parent: Puppet::Provider::Zabbix)
         macros: h['macros'].map { |macro| { macro['macro'] => macro['value'] } },
         proxy: proxy_select,
         interfacetype: interface['type'].to_i,
-        interfacedetails: interface['details']
+        interfacedetails: interface['details'],
+        tls_accept: h['tls_accept'].to_i,
+        tls_connect: h['tls_connect'].to_i,
+        tls_issuer: h['tls_issuer'].to_s,
+        tls_subject: h['tls_subject'].to_s
       )
     end
   end
@@ -60,6 +64,9 @@ Puppet::Type.type(:zabbix_host).provide(:ruby, parent: Puppet::Provider::Zabbix)
 
     proxy_hostid = @resource[:proxy].nil? || @resource[:proxy].empty? ? nil : zbx.proxies.get_id(host: @resource[:proxy])
 
+    tls_accept = @resource[:tls_accept].nil? ? 1 : @resource[:tls_accept]
+    tls_connect = @resource[:tls_connect].nil? ? 1 : @resource[:tls_connect]
+
     # Now we create the host
     zbx.hosts.create(
       host: @resource[:hostname],
@@ -76,7 +83,11 @@ Puppet::Type.type(:zabbix_host).provide(:ruby, parent: Puppet::Provider::Zabbix)
         }
       ],
       templates: templates,
-      groups: groups
+      groups: groups,
+      tls_connect: tls_connect,
+      tls_accept: tls_accept,
+      tls_issuer: @resource[:tls_issuer].nil? ? '' : @resource[:tls_issuer],
+      tls_subject: @resource[:tls_subject].nil? ? '' : @resource[:tls_subject]
     )
   end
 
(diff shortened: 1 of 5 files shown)


teluq-pbrideau:
What do you recommend for me to test my changes here?
1- Should I wait for #850 to be merged and rebase this branch from master?
2- Or is it good practice to merge from another branch directly on this PR?

lib/puppet/provider/zabbix_host/ruby.rb
```diff
@@ -86,6 +102,7 @@ def exists?
 
   def destroy
     zbx.hosts.delete(zbx.hosts.get_id(host: @resource[:hostname]))
+    @property_flush[:destroyed] = true
```
bastelfreak, line 105: awesome PR! I'm always happy to see type/provider adjustments/extensions.
I'm wondering about this change. Does it fix any broken resource cleanup / is it actually related to the TLS implementation?

teluq-pbrideau: You are right, there is no difference or without this `property_flush`.  In fact, the `destroy` function is not even possible to call in the present state of the module, I had to manually add an `ensure => absent` in the `@@zabbix_host` call in `resources/agent.pp` to test it, and it went flawlessly.

I’ve also removed the `def initialize()` that was only creating this `@property_flush`
```
  def initialize(value = {})
    super(value)
    @property_flush = {}
  end
```

Thanks for pointing this out

spec/acceptance/zabbix_host_spec.rb
```diff
@@ -308,6 +340,63 @@ class { 'zabbix':
         end
 
       end
+
+      context 'test5.example.com' do
+        let(:test5) { result_hosts.select { |h| h['host'] == 'test5.example.com' }.first }
+
+        it 'is created' do
```
teluq-pbrideau, line 347: I’ve duplicated the previously tested node, but should I skip to testing only `tls_accept`, `tls_connect`, `tls_issuer` and `tls_subject` ?

teluq-pbrideau:
Tests for CentOS 7 still broken, I’ll try to rebase my fixes and make the test run again

spec/acceptance/zabbix_host_spec.rb
```diff
@@ -132,8 +132,40 @@ class { 'zabbix':
...
+        }
+      EOS
+
+      it 'creates host with tls connections configured without errors' do
+        apply_manifest(pp4, catch_failures: true)
+      end
+
+      it 'creates host with tls connections configured without changes' do
+        apply_manifest(pp4, catch_changes: true)
+      end
```
root-expert, line 156: Although every other example is like that, we now have [this](https://github.com/voxpupuli/voxpupuli-acceptance#an-idempotent-resource) which make it little bit shorter than running `apply_manifest` 2 times

```suggestion
      it_behaves_like 'an idempotent resource' do
        let(:manifest) do
          <<-EOS
          zabbix_host { 'test5.example.com':
            ipaddress   => '127.0.0.5',
            use_ip      => false,
            port        => 1051,
            groups      => ['Virtual machines'],
            templates   => #{template},
            macros      => [],
            tls_accept  => 'cert',
            tls_connect => 'cert',
            tls_issuer  => 'Zabbix.com',
            tls_subject => 'MyClientCertificate',
          }
        EOS
```

spec/acceptance/zabbix_host_spec.rb
root-expert: I suppose some more attributes like `groups`, `macros` and `templates` are just spam and can be removed.
You can leave the basic test `is created`.

manifests/params.pp
```diff
@@ -322,6 +322,8 @@
   $agent_tlsaccept                          = undef
   $agent_tlscafile                          = undef
   $agent_tlscertfile                        = undef
+  $agent_tlscertissuer                      = undef
+  $agent_tlscertsubject                     = undef
```
root-expert, line 326: I think these can be removed and add the default value in `agent.pp` since it's the only place that they are being used.

lib/puppet/type/zabbix_host.rb
```diff
@@ -123,6 +136,36 @@ def insync?(is)
     desc 'Whether it is monitored by an proxy or not.'
   end
 
+  newproperty(:tls_connect) do
+    desc 'Connections to host.'
```
root-expert, line 140: Minor, but maybe we can add a better description here since that goes into the REFERENCE.md file. Same for `tls_accept`.

root-expert (commented):
Some minor comments here and there 😄 
CentOS specs can be ignored since that's failing on the whole orginization

spec/acceptance/zabbix_host_spec.rb
teluq-pbrideau: done, thanks

manifests/params.pp
teluq-pbrideau: I’m not sure what is the etiquette about resolving the conversations, should I resolve the conversation if the code is changed as the comment suggested, or should I let you close the conversation? :innocent:

root-expert: You can always resolve it and i'll got through it anyway 😄

spec/acceptance/zabbix_host_spec.rb
```diff
@@ -132,8 +132,36 @@ class { 'zabbix':
...
+      it_behaves_like 'an idempotent resource' do
+        let(:manifest) do
+          <<-EOS
+          zabbix_host { 'test5.example.com':
+            ipaddress   => '127.0.0.5',
+            use_ip      => false,
+            port        => 1051,
+            groups      => ['Virtual machines'],
+            templates   => #{template},
+            macros      => [],
```
root-expert, line 144: Please drop these as well, they don't add any value, just more useless API calls

root-expert: Maybe I was wrong? Revert that and it's good to go. Sorry 😄

lib/puppet/provider/zabbix_host/ruby.rb
```diff
@@ -224,4 +235,37 @@ def proxy=(string)
...
+    )
+  end
+
+  def tls_subject=(string)
+    @property_hash[:tls_subject] = string
+    zbx.hosts.create_or_update(
+      host: @resource[:hostname],
+      tls_subject: @property_hash[:tls_subject].nil? ? '' : @property_hash[:tls_subject]
+    )
+  end
```
root-expert, line 270: These could refactored to use provider flushing instead of making individual API calls, probably faster and less resources are wasted. 
A topic for another PR since more properties have to be refactored 😄

root-expert (approved):

lib/puppet/provider/zabbix_host/ruby.rb
teluq-pbrideau: Yeah, it sure sounds like a great idea, but I’m not that familiar with custom provider, not so sure if I could pull it off! :smile:

PR #851, opened 2022-11-18
use provided database port in pgpass

teluq-pbrideau:
When configuring a postgresql database that is not running on port 5432, the `.pgpass` is not correctly configured.

Then, the `exec { 'zabbix_server_create.sql':}` does not succeed as the server could not connect to the database.

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -18,7 +18,7 @@ class zabbix::database::postgresql (
   $database_user                                      = '',
   $database_password                                  = '',
   $database_host                                      = '',
-  Optional[Stdlib::Port::Unprivileged] $database_port = undef,
+  Stdlib::Port::Unprivileged $database_port           = 5432,
   $database_path                                      = $zabbix::params::database_path,
 ) inherits zabbix::params {
   assert_private()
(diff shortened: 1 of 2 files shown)


manifests/database/postgresql.pp
```diff
@@ -68,10 +68,16 @@
     }
   }
 
+  if $database_port != undef {
+    $db_port = $database_port
+  } else {
+    $db_port = '5432'
+  }
```
smortex, line 75: Maybe more readable as:

```suggestion
  $db_port = pick($database_port, '5432')
```

([from stdlib](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/REFERENCE.md#pick)) or ([using built-in functions](https://puppet.com/docs/puppet/7/function.html#lest)):

```suggestion
  $db_port = $database_port.lest || { '5432' }
```

teluq-pbrideau: True, I was simply trying to keep the same syntax from line 48 that is doing the same thing:

```
48  if $database_port != undef {
49    $port = "-p ${database_port} "
50  } else {
51    $port = ''
52  }
```

But i’ll gladly make my part more readable :+1:

smortex: Ah yes, we sometimes lack this context when reviewing :-) When facing this, I generally work on my feature in a first commit, and improve / enhance / fix style in a separate commit in the same PR :-D

teluq-pbrideau: No problem.  Would you like me to fix line 48 like so:
```
$port = $database_port ? {
  undef   => '',
  default => "-p $database_port",
 }
```
I would slightly consider it cleaner code, would you?

manifests/database/postgresql.pp
```diff
@@ -68,10 +68,12 @@
     }
   }
 
+  $db_port = pick($database_port, '5432')
```
bastelfreak, line 71: instead of the pick, I think it makes sense to set $database_port to 5432 as default value?

teluq-pbrideau: Yeah, the previously mentioned line 48 would be deleted, and the test adjusted like I pushed just now. Is it what you have in mind?

(I just figured out how the suggestion in github worked, I could have make a suggestion directly here, sorry!)

bastelfreak (approved):

teluq-pbrideau:
I’ve squashed my commits to be cleaner, feel free to ask f you want me to squash it further

bastelfreak:
na it is fine the way it is :) lets see what CI says

teluq-pbrideau:
Well, the CI will surely crash because of renaming of template as i said in https://github.com/voxpupuli/puppet-zabbix/pull/854#issuecomment-1332360269, so I’m sure I won’t have any green tick right now :smile:

smortex:
@teluq-pbrideau so, if we rebase this one on top of master it should pass :grinning: ?

From your working directory:

```
git fetch origin       # Download the latest code we have here
git rebase origin/master # Move your commits on top of the main branch
git push -f            # Send the changes (-f is required because we re-wrote history)
```

smortex (approved):

PR #850, opened 2022-11-17
New path for database schema

teluq-pbrideau:
#### Pull Request (PR) description
My attempt at fixing the database schema issue

#### This Pull Request (PR) fixes the following issues
Fixes #838

PR #854, opened 2022-11-29
Configure DirectoryIndex by default

### LINKED ISSUE #853: Set DirectoryIndex in the zabbix module

- Puppet: 7.20
- Ruby: 2.7.6p219
- Distribution: RockyLinux
- Module version: Master

```puppet
  class { 'apache':
    default_mods  => false,
  }

  class { 'zabbix' :
    [...]
  }
```

## What are you seeing
```bash
curl --head https://zabbix.example.com/
HTTP/1.1 404 Not Found
[...]
curl https://zabbix.example.com/index.php --head
HTTP/1.1 200 OK
[...]
```

I expect the zabbix web page to be available by default once configured.

Disabling the `defaults_mods` in `apache` render the zabbix web page not fully functionnal.

I included the `apache::mod::dir` in my zabbix profile, but it could probably be included by defaut in the zabbix module.  Am I wrong to think this?

teluq-pbrideau:
Make sure the DirectoryIndex is configured when apache 'default_mods` are disabled:
```puppet
  class { 'apache':
    default_mods  => false,
  }

  class { 'zabbix' :
    [...]
  }
```

Feel free to correct me if you think it should not be in this module, but instead only in my own `profile`.

Fixes #853

### DIFF
--- a/manifests/web.pp
+++ b/manifests/web.pp
@@ -306,6 +306,7 @@ class zabbix::web (
   # Is set to true, it will create the apache vhost.
   if $manage_vhost {
     include apache
+    include apache::mod::dir
     if $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '7') >= 0 and versioncmp($zabbix_version, '5') >= 0 {
       if versioncmp($facts['os']['release']['major'], '7') == 0 {
         $fpm_service = 'rh-php72-php-fpm'


smortex:
If the module is required, then it make sense to include it :+1:.

There seems to be some style issues in this file, do you mind adding a commit to fix those?

teluq-pbrideau:
the problems aren’t related to my changes, but here is the linter fixed
I expect the other tests to fail, I fixed the problem in #850 , do you recomment me to merge the changes to this PR, or to wait for it to be merged in master?

smortex:
Ah, thanks for the pointer! The other PR looked good, I merged it.

Can you please rebase your changes on top of the main branch (the second commit being already in master it should be removed in the process)?

From your working directory:

```
git fetch origin       # Download the latest code we have here
git rebase origin/master # Move your commits on top of the main branch
git push -f            # Send the changes (-f is required because we re-wrote history)
```

teluq-pbrideau:
well, the archlinux build seems broken:
```
archlinuxrolling-64.example.com 14:26:51$ puppet module install puppetlabs-postgresql -v 7.5.0
  Could not find 'english' (>= 0.a) among 97 total gem(s)
  Checked in 'GEM_PATH=/root/.local/share/gem/ruby/3.0.0:/usr/lib/ruby/gems/3.0.0' , execute `gem env` for more information
```

teluq-pbrideau:
```
Error: The template Linux SNMP does not exist in Zabbix. Please use a correct one.
```
It appears upstream has renamed the snmp templates:
`Linux SNMP` -> `Linux by SNMP`

It should probably be fixed in its own PR

smortex:
> It should probably be fixed in its own PR

Fine. You can create the feature branch on top of this branch if it helps you with testing, just remind us that "This PR also include #<this PR> " in the new PR.

The CI failure put aside, this LGTM!

smortex (approved):

smortex:
#855 has been merged, which should fix CI. The change is unlikely to have any impact regarding CI so let's merge this, thanks!

PR #856, opened 2022-12-09
Add option to restrict API access to specific hosts

teluq-pbrideau:
By default, the api is open to everyone.  I would like to restrict it to only the zabbix server.

I’m not sure if it should be included in this module like I suggest, or if I should manage my own apache config by disabling this module apache management  `manage_vhost => false` in `web.pp`, but I think the changes are simple enough to include in this module

My suggestion in this PR allow to add restriction to the api like so:
```puppet
class { 'zabbix::web' :
  [...]
  zabbix_api_access => [$facts['networking']['fqdn']],
}
```

This creates this `location` entry in apache config (or equivalent in apache 2.2):
```apache
<Location "/api_jsonrpc.php" >
  Require host zabbix.example.com
</Location>
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -74,6 +74,7 @@
 # @param ldap_reqcert Specifies what checks to perform on a server certificate
 # @param zabbix_api_user Name of the user which the api should connect to. Default: Admin
 # @param zabbix_api_pass Password of the user which connects to the api. Default: zabbix
+# @param zabbix_api_access Which host has access to the api. Default: no restriction
 # @param listenport Listen port for the zabbix-server. Default: 10051
 # @param sourceip Source ip address for outgoing connections.
 # @param logfile Name of log file.
@@ -250,6 +251,7 @@ class zabbix (
   Optional[Enum['never', 'allow', 'try', 'demand', 'hard']] $ldap_reqcert     = $zabbix::params::ldap_reqcert,
   $zabbix_api_user                                                            = $zabbix::params::server_api_user,
   $zabbix_api_pass                                                            = $zabbix::params::server_api_pass,
+  Optional[Array[Stdlib::Fqdn]] $zabbix_api_access                            = $zabbix::params::server_api_acces,
   $listenport                                                                 = $zabbix::params::server_listenport,
   $sourceip                                                                   = $zabbix::params::server_sourceip,
   Enum['console', 'file', 'system'] $logtype                                  = $zabbix::params::server_logtype,
@@ -364,6 +366,7 @@ class zabbix (
     apache_listenport_ssl                    => $apache_listenport_ssl,
     zabbix_api_user                          => $zabbix_api_user,
     zabbix_api_pass                          => $zabbix_api_pass,
+    zabbix_api_access                        => $zabbix_api_access,
     database_host                            => $database_host,
     database_name                            => $database_name,
     database_schema                          => $database_schema,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -251,7 +251,7 @@ class zabbix (
   Optional[Enum['never', 'allow', 'try', 'demand', 'hard']] $ldap_reqcert     = $zabbix::params::ldap_reqcert,
   $zabbix_api_user                                                            = $zabbix::params::server_api_user,
   $zabbix_api_pass                                                            = $zabbix::params::server_api_pass,
-  Optional[Array[Stdlib::Fqdn]] $zabbix_api_access                            = $zabbix::params::server_api_acces,
+  Optional[Array[Stdlib::Host,1]] $zabbix_api_access                          = $zabbix::params::server_api_access,
   $listenport                                                                 = $zabbix::params::server_listenport,
   $sourceip                                                                   = $zabbix::params::server_sourceip,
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -250,6 +251,7 @@
   Optional[Enum['never', 'allow', 'try', 'demand', 'hard']] $ldap_reqcert     = $zabbix::params::ldap_reqcert,
   $zabbix_api_user                                                            = $zabbix::params::server_api_user,
   $zabbix_api_pass                                                            = $zabbix::params::server_api_pass,
+  Optional[Array[Stdlib::Fqdn]] $zabbix_api_access                            = $zabbix::params::server_api_acces,
```
alexjfisher, line 254: Agree with @root-expert that `Stdlib::Host` might be a better choice.  You've also got a typo at the end of the line (a missing `s` in `access`).
```suggestion
Optional[Array[Stdlib::Host],1] $zabbix_api_access                          = $zabbix::params::server_api_access,
```

manifests/web.pp
```diff
@@ -400,6 +402,15 @@
       $directory_deny = { 'deny' => 'from all', 'order' => 'Deny,Allow', }
     }
 
+    $location_api_access = $zabbix_api_access ? {
+      undef   => $directory_allow,
+      default => if versioncmp($apache::apache_version, '2.4') >= 0 {
```
bastelfreak, line 407: I'm wondering if we really need to support apache 2.2. dropping this logic makes the whole file way easier to read.

alexjfisher: Yeah, a separate commit to drop that makes sense I think.  After all, are there any supported OSes that still use Apache 2.2??
If this was zabbix-agent code then I'm all for being more conservative in what we drop, but for the server side stuff, not so much.

FWIW, I had a think about [constructs](https://gist.github.com/alexjfisher/eeda5a8d54f9895f37b7b3aafea54d6e) that might be more readable.  Conclusion was they're all pretty horrible!

teluq-pbrideau: Do you think about removing entirely the `$directory_allow` / `$directory_deny` variables, thus removing an entire `merge()`, like so:
```diff
-        merge(
-          merge({
-              path     => '/usr/share/zabbix',
-              provider => 'directory',
-          }, $directory_allow),
-          $fcgi_filematch
+        merge({
+            path     => '/usr/share/zabbix',
+            provider => 'directory',
+            require  => 'all granted',
+          }, $fcgi_filematch

```

Or should I keep the variables, just remove the support for old apache like so:
```diff
-    # Check which version of Apache we're using
-    if versioncmp($apache::apache_version, '2.4') >= 0 {
-      $directory_allow = { 'require' => 'all granted', }
-      $directory_deny = { 'require' => 'all denied', }
-    } else {
-      $directory_allow = { 'allow' => 'from all', 'order' => 'Allow,Deny', }
-      $directory_deny = { 'deny' => 'from all', 'order' => 'Deny,Allow', }
-    }
-
+    $directory_allow = { 'require' => 'all granted', }
+    $directory_deny = { 'require' => 'all denied', }
     $location_api_access = $zabbix_api_access ? {
       undef   => $directory_allow,
-      default => if versioncmp($apache::apache_version, '2.4') >= 0 {
-        { 'require' => $zabbix_api_access.map |$host| { "host ${host}" }, }
-      } else {
-        { 'allow' => $zabbix_api_access, 'order' => 'Allow,Deny' }
-      }
+      default => { 'require' => $zabbix_api_access.map |$host| { "host ${host}" }},
     }
```

And keep the mess down?
```
        merge(
          merge({
              path     => '/usr/share/zabbix',
              provider => 'directory',
          }, $directory_allow),
          $fcgi_filematch
```

teluq-pbrideau: I did option 1, but tell me if you think it should be done otherwise.

teluq-pbrideau:
Or should I make a less precise test like what is done in `apache` module:
```spec
            is_expected.to contain_concat__fragment('zabbix.example.com-directories')
              .with_content(%r{^\s+Require host 127.0.0.1$})
```

manifests/web.pp
teluq-pbrideau: Aslo, I added an explicit `fail()` for apache < 2.4

root-expert:
> Or should I make a less precise test like what is done in `apache` module:
> 
> ```python
>               it {
>                 is_expected.to contain_concat__fragment('zabbix.example.com-directories').with(
>                   content: %r{^\s+Require host 127.0.0.1$},
>                 )
>               }
> ```

Although the first option is better that's also okay for me. Probably you need something like this though:

```ruby
               it {
                 is_expected.to contain_concat__fragment('zabbix.example.com-directories').with(
                   content: %r{^\s*Require host 127\.0\.0.\1$},
                 )
               }
```
Dot matches everything except new lines, so you need to escape it 😄

teluq-pbrideau:
Note: The debian11 test has crashed out of my scope
```
  Error: Could not set 'present' on ensure: Connection reset by peer - SSL_connect (file: /etc/puppetlabs/code/environments/production/modules/apt/manifests/key.pp, line: 54)
```

root-expert (approved):

root-expert:
@teluq-pbrideau Can you please rebase with our master branch one last time before merging this?

PR #857, opened 2022-12-12
Duplicate: Add sensitive support

### LINKED ISSUE #440: Manage resources leaks API password in resource types

### LINKED ISSUE #950: Accept Datatype Sensitive[String] for Passwords
To not reveal sensitive Data accept Datatype `Sensitive[String]` for Passwords.  

Motivation for this is:  Authors of Profilemodules should not be bothered with dealing with Datatype `Sensitive`.  Componentmodules (like `zabbix`) should take Care of this.

Examples:  
```puppet
class zabbix (
  Variant[Sensitive[String], String] $database_password = $zabbix::params::server_database_password,
) {}
```

```puppet
class zabbix::database (
  Variant[Sensitive[String], String] $database_password = $zabbix::params::server_database_password,
) {}
```

teluq-pbrideau:
This is a copy of #828, where @bdeferme do not have any more time to work on it.  Here I tried to fix the tests.

Fixes #440
Fixes #950


teluq-pbrideau:
The fail on Debian 10 appears to be unrelated to this change:
```
  Error: /Stage[main]/Zabbix::Web/Package[zabbix-frontend-php]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install zabbix-frontend-php' returned 100: Reading package lists...
  Building dependency tree...
  Reading state information...
  The following NEW packages will be installed:
    zabbix-frontend-php
  0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
  Need to get 2559 kB of archives.
  After this operation, 18.1 MB of additional disk space will be used.
  Err:1 http://repo.zabbix.com/zabbix/4.0/debian buster/main amd64 zabbix-frontend-php all 1:4.0.44-1+buster
    Connection failed [IP: 178.128.6.101 80]
  E: Failed to fetch http://repo.zabbix.com/zabbix/4.0/debian/pool/main/z/zabbix/zabbix-frontend-php_4.0.44-1+buster_all.deb  Connection failed [IP: 178.128.6.101 80]
  E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

teluq-pbrideau:
While trying to add some basic tests about this change, I’ve discovered that the `zabbix::proxy` tests aren’t run at all, it is restricted to `when 'RedHat'` which is not in the listed OS `on_supported_os(baseline_os_hash)`…

I’ve fixed the tests for them to run at least for CentOS, but they fail for other OS for reasons outside the scope of this PR:
```
zabbix::proxy
  on debian-10-x86_64
[...]
      is expected to contain Yumrepo[zabbix-nonsupported] (FAILED - 1)
```
It should probably be fixed it its own PR…

manifests/params.pp
```diff
@@ -182,7 +182,7 @@
   $ldap_clientcert                          = undef
   $ldap_clientkey                           = undef
   $ldap_reqcert                             = undef
-  $server_api_pass                          = 'zabbix'
+  $server_api_pass                          = Sensitive('zabbix')
```
teluq-pbrideau, line 185: I’m not sure if we must set default values as sensitive.

I’ve changed some tests because of this: if the default value is `Sensitive`, the output may be masked during tests, and the validation cannot be made.  These lines are marked with this comment below:

```
# cleartext password must be explicitly declared in this test, otherwise the parser will secure content of the file
```

For example, see `spec/classes/proxy_spec.rb:115`

manifests/server.pp
```diff
@@ -66,6 +66,9 @@
 # @param startreportwriters Number of pre-forked report writer instances.
 # @param webserviceurl URL to Zabbix web service, used to perform web related tasks.
 # @param vmwarefrequency How often zabbix will connect to vmware service to obtain a new datan.
+# @param vmwareperffrequency
+#   Delay in seconds between performance counter statistics retrieval from a single VMware service.
+#   This delay should be set to the least update interval of any VMware monitoring item that uses VMware performance counters.
```
root-expert, line 71: It would be better to open a new PR for the new parameter

manifests/server.pp
```diff
@@ -132,6 +135,7 @@
 # @param include_dir You may include individual files or all files in a directory in the configuration file.
 # @param loadmodulepath Full path to location of server modules.
 # @param loadmodule Module to load at server startup.
+# @param sslcalocation_dir Location of certificate authority (CA) files for SSL server certificate verification.
```
root-expert, line 138: It would be better to open a new PR for the new parameter

manifests/params.pp
root-expert: IMO that's okay

root-expert (commented):
Thanks for the PR! I left some inline comments :smile:

manifests/server.pp
teluq-pbrideau: My guess is @bdeferme discovered the following parameters in `zabbix_server.conf` while converting to the new epp template:

- SSLCALocation
- VMwarePerfFrequency

As they were present in upstream config, and noticed they were missing in the old erb template, and added them to be consistent with upstream config.

@root-expert  I’ve deleted them as requested, but I think they should be included.  However, I won’t go out of my way to create a new PR if it is not required in my own production environment.

bastelfreak:
@teluq-pbrideau thanks for working on this! Can you please rebase against our master branch?

teluq-pbrideau:
@bastelfreak I’m sorry but rebasing was just a pain in the ass with all the conflicts. I solved the conflict once with the merge of your master branch

teluq-pbrideau:
You might see I added back the `SSLCALocation` and `VMwarePerfFrequency` parameters as they were already defined in the `.erb` template, without beeing defined in the class.  I think it is cleaner to follow how the template was, and adapting the class accordingly.  It is such a pain to modify and update the erb vs the epp to keep them consistent.

teluq-pbrideau:
I’m not quite sure how to cleanup my git tree…  Can we just squash to only one commit when we merge into master?

teluq-pbrideau:
I’ve been running the old changes in my production environment since my original PR, but I cannot test these new changes right now in a live environment.  I’m stuck with a dependency cycle in my `Puppetfile` with `stdlib9`, which I’m in the process of solving right now, but might take a few days still to solve on my side.
Feel free to freeze this PR until I come back with confirmation on my side, or if someone can test in a more complete environment.

teluq-pbrideau:
I’ve just successfully tested this PR in my development environment.  The failing tests are out of my control, it was a timeout after 360 minutes:
```
Error: The operation was canceled.
```
Can we force the tests to rerun and move forward with this? Or It can also be the changes #973!

teluq-pbrideau:
Hoping #973 will be merged someday…

PR #859, opened 2023-01-09
Remove check for Apache 2.4

JoshuaSmeda:
Apache 9.0.0 release [drops](https://github.com/puppetlabs/puppetlabs-apache/pull/2329) 2.2 support. Since Zabbix has a dependency on Apache, it makes sense to remove the comparison between >= `2.4` and `2.2`, including setting the minimum dependency version for Apache to 9.0.0.

The Zabbix module breaks if you run Apache version higher than `9.0.0`, due to not being able to verify the fact.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -18,7 +18,7 @@
     },
     {
       "name": "puppetlabs/apache",
-      "version_requirement": ">= 8.0.0 < 9.0.0"
+      "version_requirement": ">= 9.0.0"
     },
     {
       "name": "puppetlabs/firewall",
--- a/metadata.json
+++ b/metadata.json
@@ -18,7 +18,7 @@
     },
     {
       "name": "puppetlabs/apache",
-      "version_requirement": ">= 9.0.0"
+      "version_requirement": ">= 9.0.0 < 10.0.0"
     },
     {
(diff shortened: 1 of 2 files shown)


smortex (approved):
LGTM

metadata.json
```diff
@@ -18,7 +18,7 @@
     },
     {
       "name": "puppetlabs/apache",
-      "version_requirement": ">= 8.0.0 < 9.0.0"
+      "version_requirement": ">= 9.0.0"
```
root-expert, line 21: Maybe we should cap it below 10 as well?
```suggestion
      "version_requirement": ">= 9.0.0 < 10.0.0"
```

smortex: Wow, I got distracted :-) Indeed we should do that!

JoshuaSmeda: Committed, thanks for review.

bastelfreak:
I think it's fine to to mark this backwards-incompatible. We deprecated CentOS 6 in the past, and that's the only OS that used Apache 2.2.

evgeni:
While dropping C6 support is fine, bumping puppetlabs/apache is backwards incompatible.

PR #864, opened 2023-02-28
Update repo.pp

SK4ndal:
Update gpg key as per comments on stale PR 856
Initial PR by [Joshua-Snapp](https://github.com/voxpupuli/puppet-zabbix/issues?q=is%3Apr+is%3Aopen+author%3AJoshua-Snapp)

### DIFF
--- a/lib/puppet/type/zabbix_host.rb
+++ b/lib/puppet/type/zabbix_host.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 Puppet::Type.newtype(:zabbix_host) do
+  desc 'Zabbix host type'
   ensurable do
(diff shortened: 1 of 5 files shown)


SK4ndal:
Why is the pipeline failing? I rebased on master and it says its up to date. 
Is this also broken in Master? 
I don't speak ruby. Does this check mean that we need to add "desc" statements like in [these other statements](https://github.com/voxpupuli/puppet-zabbix/blob/master/lib/puppet/type/zabbix_host.rb#L44)

SK4ndal:
Ok I tried some things but they don't seem to work. I'd be very gratefull if someone could advice me.

PR #867, opened 2023-03-22
Add support for FreeBSD only for Zabbix Agent

gizmo15:
Add support for FreeBSD only for Zabbix Agent.

Tested on FreeBSD 13.1 with puppet 7.23


fraenki:
@gizmo15 Would you please rebase against current master branch?

evgeni (approved):

PR #874, opened 2023-04-16
modulesync 7.0.0

maxadamo:


evgeni:
The only failures are from the "new" OSes which were not tested before. I am inclined to merging this as-is and then following up with fixes for Oracle.

manifests/repo.pp
```diff
@@ -87,7 +87,7 @@
           }
         }
 
-        if ($facts['os']['release']['major'] == '7' and versioncmp($zabbix_version, '5.0') >= 0) {
+        if ($facts['os']['name'] == 'CentOS' and $facts['os']['release']['major'] == '7' and versioncmp($zabbix_version, '5.0') >= 0) {
```
ekohl, line 90: Will this also affect RHEL? If you're running RHEL, you probably don't want CentOS SCL bits, but it can break some environments.

ekohl (approved):
Normally I squash the modulesync commits when I can so you don't see intermediate versions when you don't need to, but I won't hold it back on that.

manifests/repo.pp
ekohl: Oh yes, with Foreman we had copied it in our repos so it worked for other OSes (IIRC mostly for Scientific Linux).

evgeni: Those were the times…

PR #875, opened 2023-05-17
Restarting agent if it get's upgraded

hertell:
I recently upgraded all agents to the latest 6.4 version of zabbix, and stumbled into Zabbix having issues eg getting old agents not compatible with new items from 6.4. I figured out that the upgrade of the agent-package did upgrade the agent, but it did not restart them, which left the old agent running. This had  me to restart all agents manually. This fix will restart the agent if it get's upgraded.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -347,9 +347,11 @@ class zabbix::agent (
 
   # Controlling the 'zabbix-agent' service
   service { $servicename:
-    ensure  => $service_ensure,
-    enable  => $service_enable,
-    require => $service_require,
+    ensure    => $service_ensure,
+    enable    => $service_enable,
+    require   => $service_require,
+    subscribe => Package[$zabbix_package_agent],
+
   }
 
   # Override the service provider on AIX


manifests/agent.pp
```diff
@@ -347,9 +347,11 @@
 
   # Controlling the 'zabbix-agent' service
   service { $servicename:
-    ensure  => $service_ensure,
-    enable  => $service_enable,
-    require => $service_require,
+    ensure    => $service_ensure,
+    enable    => $service_enable,
+    require   => $service_require,
+    subscribe => Package[$zabbix_package_agent],
```
evgeni, line 353: This should already be present via the `$service_require`, which is defined as:

```
  $service_require = $manage_startup_script ? {
    true  => [Package[$zabbix_package_agent], Zabbix::Startup[$servicename]],
    false => Package[$zabbix_package_agent]
  }
```

If something requires a thing, it will get notified about changes.

evgeni: Do you set `zabbix_package_state => latest`? Otherwise Puppet would not have updated the package and nothing would have issued the notification.

PR #876, opened 2023-06-06
getting latest puppet-agent in place of the old 4.4.5 for Windows

tunisiano187:
Updating to latest instead of 4.4.5


ekohl (changes requested):
You're only adjusting the tests, but that doesn't modify the actual code.

PR #878, opened 2023-07-01
Allow puppetlabs/stdlib 9.x, puppetlabs/postgresql 9.x, puppetlabs/mysql 15.x, pyppetlabs/apache 11.x, puppetlabs/apt 9.x, puppet/systemd 6.x, puppet/selinux 4.x

bastelfreak:


Josephvh:
@bastelfreak any idea when this could be released to Master? I'd like to create a PR for releasing version 10.0.0 but would like the option of upgrading to stdlib 9+ so I can get my entire env over (especially with all the latest modules requiring a min on stdlib 9)

kenyon:
@Josephvh this needs work to get the test passing, and a modulesync in #874.

ekohl (approved):

PR #884, opened 2023-09-19
Fixed missing item when syncing templates

Fabian1976:


root-expert:
Please update REFERENCE.md to allow the CI continue:

```
bundle exec rake strings:generate:reference
```

Fabian1976:
I haven't forgotten this. I'm trying to get my ruby/rake shizzle to work. We don't use it anymore (we use pre-commit and do a lot of this stuff in our pipelines), but i will try to get it to work again.

PR #885, opened 2023-09-19
Fix java-gateway pid different in zabbix 6

Fabian1976:


root-expert (changes requested):
Left some comments, let me know if it makes sense :smile: 

Please rebase with our master branch as well.

manifests/javagateway.pp
```diff
@@ -24,6 +24,11 @@
   $startpollers             = $zabbix::params::javagateway_startpollers,
   $timeout                  = $zabbix::params::javagateway_timeout,
 ) inherits zabbix::params {
+  # Fix for pid file. Is different in Zabbix (4, 5) and 6
+  $real_pidfile = $zabbix_version ? {
+    /^[45]\.[024]/ => '/var/run/zabbix/zabbix_java.pid',
+    /^[6]\.[024]/  => $pidfile,
+  }
```
root-expert, line 31: This hardcodes the PID file for 4.x and 5.x. Shouldn't the user be able to change the path? 
Maybe we can use the `pick()` function and set the default value of `javagateway_pidfile` to `undef`

```suggestion
  # Fix for pid file. Is different in Zabbix (4, 5) and 6
  $real_pidfile = $zabbix_version ? {
    /^[45]\.[024]/ => pick($pidfile, '/var/run/zabbix/zabbix_java.pid'),
    /^[6]\.[024]/  => pick($pidfile, '/var/run/zabbix/zabbix_java_gateway.pid)
  }
```

manifests/params.pp
```diff
@@ -472,7 +472,7 @@
   # Java Gateway specific params
   $javagateway_listenip                     = '0.0.0.0'
   $javagateway_listenport                   = '10052'
-  $javagateway_pidfile                      = '/var/run/zabbix/zabbix_java.pid'
+  $javagateway_pidfile                      = '/var/run/zabbix/zabbix_java_gateway.pid'
```
root-expert, line 475:
```suggestion
  $javagateway_pidfile                      = undef
```

manifests/javagateway.pp
Fabian1976: Yes, agreed. This is a much nicer solution.

manifests/params.pp
Fabian1976: Also updated this part.

root-expert (approved):

PR #888, opened 2023-10-03
Drop Debian 10 and Ubuntu 18.04 support

evgeni:


ekohl (approved):
https://learn.microsoft.com/en-us/lifecycle/announcements/windows-server-2012-r2-end-of-support suggests that in 10 days Windows 2012 and 2012 R2 are also EOL. Perhaps drop those too? :)

And Raspbian also looks outdated.

PR #890, opened 2023-10-04
remove more Zabbix 5.4 special cases now that it's unsupported

evgeni:


spec/classes/database_mysql_spec.rb
```diff
@@ -29,8 +29,6 @@
         # path to sql files on Debian and RedHat
         path = if Puppet::Util::Package.versioncmp(zabbix_version, '6.0') >= 0
                  '/usr/share/zabbix-sql-scripts/mysql/'
-               elsif Puppet::Util::Package.versioncmp(zabbix_version, '5.4') >= 0
-                 '/usr/share/doc/zabbix-sql-scripts/mysql/'
                else
                  '/usr/share/doc/zabbix-*-mysql*'
```
bastelfreak, line 33: I assume those defaults are valid for the LTS zabbix?

bastelfreak (approved):

PR #892, opened 2023-10-04
don't call puppet multiple times in prepare host for acceptance

evgeni:


bastelfreak (approved):

evgeni:
this speeds up the cleanup from ~20sec to ~5 sec :)

PR #893, opened 2023-10-04
add EL9 support

evgeni:


root-expert (approved):

bastelfreak (approved):

evgeni:
* Switching to 6.0 by default is in #894 
* Given pre-6.0 is not available on EL9 (besides agent), should the other classes `fail()` on those combos?

bastelfreak (approved):

PR #895, opened 2023-10-04
increase default CacheSize to 32M, the default in 6.0 

evgeni:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -194,7 +194,7 @@ class zabbix::params {
   # Zabbix-server
   $server_alertscriptspath                  = '/etc/zabbix/alertscripts'
   $server_allowroot                         = '0'
-  $server_cachesize                         = '8M'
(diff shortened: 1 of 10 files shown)


manifests/params.pp
```diff
@@ -194,7 +194,7 @@
   # Zabbix-server
   $server_alertscriptspath                  = '/etc/zabbix/alertscripts'
   $server_allowroot                         = '0'
-  $server_cachesize                         = '8M'
+  $server_cachesize                         = '32M'
```
evgeni, line 197: this is technically a breaking change (as it changes a default), should it be split into an own PR that is labeled `breaking`?

evgeni: we could also alternatively configure it in the tests, but then users get a broken-by-default 6.0 install which just doesn't start with a 8M of cache

bastelfreak (approved):

PR #898, opened 2023-10-05
puppetlabs/firewall: Require 7.x

evgeni:


evgeni:
Yet another of those annoying changes.

Up to 6.0 there was an `action` parameter and a `jump` parameter, which both resulted in `-J` to `iptables`, but `jump` was forbidden to be `accept`, `reject` or `drop`:
https://github.com/puppetlabs/puppetlabs-firewall/blob/v6.0.0/lib/puppet/type/firewall.rb#L674-L679
In 7.0 the `action` parameter is gone in favor of `jump`.

bastelfreak (approved):

PR #899, opened 2023-10-05
drop pidfile_workaround

evgeni:


evgeni:
The ubuntu part certainly can go, as we don't support 16.04 anymore.
Let's see what CentOS thinks.

zilchms (approved):

PR #900, opened 2023-10-05
manifests/proxy.pp: avoid undefined variable for Puppet 8 compat

spicyprogramming:

If we don't specify the `$listenip` parameter for the zabbix proxy class its value is, by default, `undef`, which in turn causes the variable `$listen_ip` to be undefined:

https://github.com/voxpupuli/puppet-zabbix/blob/58b7688015b4b0e1a3332a3892eea8fe4a186b2a/manifests/proxy.pp#L310-L323

and here `$listen_ip` is used without being ever defined:

https://github.com/voxpupuli/puppet-zabbix/blob/58b7688015b4b0e1a3332a3892eea8fe4a186b2a/manifests/proxy.pp#L336-L343

Puppet 8 seems to be more strict than puppet 7 and errors out with the following message:

```shell
Error: Unknown variable: 'listen_ip'
```

This is how i'm using the proxy class

```ruby
class { 'zabbix::proxy':
  zabbix_version       => $version,
  zabbix_package_state => 'installed',
  manage_resources     => true,
  mode                 =>  1,
  zabbix_server_host   => $server_address,
  database_type        => 'postgresql',
}
```

Without this change i can't use this module to configure the proxy on a Linux node with puppet-agent v7.26.0 connected to a puppet-server v8.2.3:

```shell
root@lazp-pupprx-01:~# puppet agent -t 
Info: Using environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Unknown variable: 'listen_ip'. (file: /etc/puppetlabs/code/environments/production/modules/zabbix/manifests/proxy.pp, line: 342, column: 22) on node lazp-pupprx-01
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
root@lazp-pupprx-01:~# 
```

### DIFF
--- a/manifests/proxy.pp
+++ b/manifests/proxy.pp
@@ -317,6 +317,8 @@ class zabbix::proxy (
     } else {
       $listen_ip = undef
     }
+  } else {
+    $listen_ip = undef
   }
 
   # So if manage_resources is set to true, we can send some data


smortex (approved):
`$listenip` and `$listen_ip` are very similar names.  Future refactoring might improve the naming to make them more distinguishable.

The code also seems to `getvar()` legacy facts, which will break at some point in the future.

Ignoring these points, that looks good.

PR #901, opened 2023-10-20
gha-puppet v2

evgeni:

### DIFF
--- a/spec/support/acceptance/supported_versions.rb
+++ b/spec/support/acceptance/supported_versions.rb
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
 def supported_versions
-  %w[5.0 6.0]
+  supported_versions = %w[5.0 6.0]
+  # this is a hack so that we don't have to rewrite the existing acceptance tests
+  if (beaker_zabbix_version = ENV.fetch('BEAKER_FACTER_zabbix_version', nil))
+    supported_versions &= [beaker_zabbix_version]
+    raise "Unsupported version: #{beaker_zabbix_version}" if supported_versions.empty?
+  end
+  supported_versions
 end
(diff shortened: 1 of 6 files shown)


spec/support/acceptance/supported_versions.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
 def supported_versions
-  %w[5.0 6.0]
+  # this is a hack so that we don't have to rewrite the other tests
+  [ENV.fetch('BEAKER_FACTER_ZABBIX_VERSION')]
```
ekohl, line 5: It uses the env var verbatim, so it's 
```suggestion
  [ENV.fetch('BEAKER_FACTER_zabbix_version')]
```

We can debate whether it should make it case insensitive and always uppercase the env var while lower casing the fact, but it's not here today.

evgeni: just realized the same when running m2gha locally.

spec/support/acceptance/supported_versions.rb
```diff
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
 def supported_versions
-  %w[5.0 6.0]
+  supported_versions = %w[5.0 6.0]
+  # this is a hack so that we don't have to rewrite the existing acceptance tests
+  if (beaker_zabbix_version = ENV.fetch('BEAKER_FACTER_zabbix_version', nil))
+    supported_versions &= beaker_zabbix_version
```
ekohl, line 7:
```
 Failure/Error: supported_versions &= beaker_zabbix_version
TypeError:
  no implicit conversion of String into Array
```
```suggestion
    supported_versions &= [beaker_zabbix_version]
```

ekohl (approved):

PR #903, opened 2023-11-06
move zabbixapi gem installation into own class

evgeni:

### DIFF
--- a/manifests/zabbixapi.pp
+++ b/manifests/zabbixapi.pp
@@ -1,0 +1,33 @@
+# @summary This will install the zabbixapi gem.
+#
+# @param zabbix_version This is the zabbix version.
+# @param puppetgem Provider for the zabbixapi gem package.
+class zabbix::zabbixapi (
+  $zabbix_version = $zabbix::params::zabbix_version,
+  $puppetgem = $zabbix::params::puppetgem,
+) inherits zabbix::params {
+  # Determine correct zabbixapi version.
+  case $zabbix_version {
+    /^[56]\.[024]/: {
+      $zabbixapi_version = '5.0.0-alpha1'
+    }
+    default: {
+      fail("Zabbix ${zabbix_version} is not supported!")
+    }
+  }
+
+  $compile_packages = $facts['os']['family'] ? {
+    'RedHat' => ['make', 'gcc-c++', 'rubygems', 'ruby'],
+    'Debian' => ['make', 'g++', 'ruby-dev', 'ruby', 'pkg-config',],
+    default  => [],
+  }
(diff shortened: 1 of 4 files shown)


manifests/zabbixapi.pp
```diff
@@ -0,0 +1,33 @@
...
+      fail("Zabbix ${zabbix_version} is not supported!")
+    }
+  }
+
+  $compile_packages = $facts['os']['family'] ? {
+    'RedHat' => ['make', 'gcc-c++', 'rubygems', 'ruby'],
+    'Debian' => ['make', 'g++', 'ruby-dev', 'ruby', 'pkg-config',],
+    default  => [],
+  }
+  ensure_packages($compile_packages)
```
bastelfreak, line 24: that is deprecated, please use the newer version:
```suggestion
  stdlib::ensure_packages($compile_packages)
```

evgeni: can I use this now w/o bumping stdlib to >= 9?

ekohl: I don't think so

manifests/zabbixapi.pp
```diff
@@ -0,0 +1,33 @@
...
+# @param zabbix_version This is the zabbix version.
+# @param puppetgem Provider for the zabbixapi gem package.
+class zabbix::zabbixapi (
+  $zabbix_version = $zabbix::params::zabbix_version,
+  $puppetgem = $zabbix::params::puppetgem,
+) inherits zabbix::params {
+  # Determine correct zabbixapi version.
+  case $zabbix_version {
+    /^[56]\.[024]/: {
+      $zabbixapi_version = '5.0.0-alpha1'
```
ekohl, line 12: Note to self: https://rubygems.org/gems/zabbixapi does only list a 5.0.0-alpha1 from 2021.

manifests/zabbixapi.pp
```diff
@@ -0,0 +1,33 @@
+# @summary This will install the zabbixapi gem.
+#
+# @param zabbix_version This is the zabbix version.
+# @param puppetgem Provider for the zabbixapi gem package.
+class zabbix::zabbixapi (
+  $zabbix_version = $zabbix::params::zabbix_version,
+  $puppetgem = $zabbix::params::puppetgem,
```
ekohl, line 7: Should this be `String[1]` for both?

evgeni: In theory, yes.
In reality, we do not enforce these anywhere else, so I find it odd to do it here.

(read: we should do it globally, but not in this PR)

PR #905, opened 2023-11-06
Better Hiera compatibility

ekohl:
This does the required includes for `apache` and `postgresql::server` if needed. It then uses this together with beaker-hiera (which voxpupuli-acceptance already automatically sets up for us) to reduce duplication in acceptance tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,8 +138,6 @@ node 'zabbix.example.com' {
   }
   include apache::mod::php
 
-  class { 'postgresql::server': }
-
   class { 'zabbix':
     zabbix_url    => 'zabbix.example.com',
   }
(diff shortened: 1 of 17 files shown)


bastelfreak (approved):

ekohl:
Needs https://github.com/voxpupuli/voxpupuli-acceptance/pull/77 for Puppet 8.

smortex (approved):
LGTM!  I am not sure about the extra parameters that where passed to `postgresql::globals` (but the tests seems to prove that they are not required), and including the required code is IMHO the right thing to do!

evgeni (approved):

PR #910, opened 2023-11-08
use www-data on all Debian family systems, not only on Debian and Ubuntu

evgeni:


evgeni:
> Isn't that a bugfix?

Someones bugfix is someone else's breaking change.

(You can see it in various ways)

PR #912, opened 2023-11-08
drop last bits of Fedora support

evgeni:
Fixes: cbf95846e294aca51a3b179994d8c9ae801bd500


bastelfreak (approved):

evgeni:
Is it really breaking? We've not been advertising Fedora support for a while.

(Not that it matters, the release will be *so* breaking anyway)

PR #913, opened 2023-11-08
use os.family fact to detect params

evgeni:
contains #912


evgeni:
I *think* this is backwards compatible, as with Fedora removed from that list and Debian/Ubuntu updated to Debian, everything else is actually a family. And for people who did use other OSes it's not breaking as they had to set the params themself anyway

bastelfreak (approved):

PR #869, opened 2023-03-29
Add Zabbix parameter StatsAllowedIP  for Server and Proxy and remove ServerPort for proxy in >6.0

geirra:
#### Pull Request (PR) description
This adds the optional parameter StatsAllowedIP to zabbix server and proxy.
The paramter has been supported since v4.2

#### This Pull Request (PR) fixes the following issues
n/a

PR #921, opened 2024-02-16
Add and remove Zabbix 6.0 + 6.4 options (HA mode, StatsAllowedIP, ProxyConfigFrequency, HeartbeatFrequency)

geirra:
This PR adds some new 6.0 + 6.4 functions and removes config values that have been Deprecated.

1. Add missing parameter StatsAllowedIP  for Zabbix Server + Proxys
This is a rebase of an earlier attempt #869  
The parameter has been supported since version 4.2
2.  HA Mode for Zabbix:
This adds HANodeName + NodeAddress to Zabbix server.
3. Since Zabbix 6.0 and the support for HA setup of Zabbix, the webconfig does not require `$ZBX_SERVER_PORT` and `$ZBX_SERVER `. In cases where HA is used these values have to be removed.
Offical Zabbix deployments from 6.0 has these values disabled by default, the connection details is now pulled by DB.
4. As of Zabbix 6.4 ConfigFrequency has been Deprecated and has been replaced with ProxyConfigFrequency, so the same value is now used in Server and Proxy. 
```
### Option: ConfigFrequency - Deprecated, use ProxyConfigFrequency
#	How often proxy retrieves configuration data from Zabbix Server in seconds.
#	For a proxy in the passive mode this parameter will be ignored.
# Mandatory: no

This is not to be confused with the existing server parameter ProxyConfigFrequency:
``` 
**From Zabbix wiki**
``` 
Note that ProxyConfigFrequency is:

the server parameter for passive proxies;
the proxy parameter for active proxies.
``` 
5. HeartbeatFrequency has been removed from 6.4 and up 
```
The heartbeat sender has been removed from the proxy. Therefore, Zabbix proxy item zabbix [process,heartbeat sender] is no longer supported and has been removed from templates. The HeartbeatFrequency parameter has been deprecated.
```


geirra:
Seem like this one should be fine now.

The failed tests are due to broken postgresql gpg keys, needs to fixed upstream.

`  Public key for postgresql13-13.14-1PGDG.rhel7.x86_64.rpm is not installed
  Importing GPG key 0x442DF0F8:
   Userid     : "PostgreSQL RPM Building Project <pgsqlrpms-hackers@pgfoundry.org>"
   Fingerprint: 68c9 e2b9 1a37 d136 fe74 d176 1f16 d2e1 442d f0f8
   From       : /etc/pki/rpm-gpg/RPM-GPG-KEY-PGDG-13`
   
   Probably related to https://yum.postgresql.org/news/pgdg-rpm-repo-gpg-key-update/

Valantin (approved):
Configuration checked in the official documentation, seems to be all ok, I've test it in our infrastructure as soon as possible

manifests/params.pp
```diff
@@ -367,6 +370,7 @@
   $proxy_cachesize                          = '32M'
   $proxy_configfile_path                    = '/etc/zabbix/zabbix_proxy.conf'
   $proxy_configfrequency                    = '3600'
+  $proxy_proxyconfigfrequency               = undef
```
root-expert, line 373: According to the documentation the default value is 10

```suggestion
  $proxy_proxyconfigfrequency               = 10
```

manifests/server.pp
```diff
@@ -268,12 +271,16 @@
   $loadmodule                                                                 = $zabbix::params::server_loadmodule,
   $sslcertlocation_dir                                                        = $zabbix::params::server_sslcertlocation,
   $sslkeylocation_dir                                                         = $zabbix::params::server_sslkeylocation,
+  Optional[String[1]] $statsallowedip                                         = $zabbix::params::server_statsallowedip,
+
```
root-expert, line 275:
```suggestion
```

geirra:
So this sould be rebased and the empty line is removed, i have not fixed the issue about ProxyConfigFrequency, see the comment about that one.

Mind taking a look  @root-expert ?

root-expert (approved):

PR #914, opened 2023-11-13
Bootstrap agent2 support (rebased)

Phil-Friderici:
In this PR I try to continue the work of fe80 (and others) in [PR824](https://github.com/voxpupuli/puppet-zabbix/pull/824).

#### Pull Request (PR) description
This Pr bootstrap a support for zabbix-agent2

* Add param binary_location for specify zabbix agent binary (only for systemd)
* Remove some parameters in zabbix agent configuration if it's zabbix agent2 package
  * LogRemoteCommands
  * StartAgents
  * MaxLinesPerSecond
  * AllowRoot
  * User
  * LoadModulePath

It's possible to use agent2 with this configration:
```
      'agent_configfile_path' => '/etc/zabbix/zabbix_agent2.conf',
      'include_dir'           => '/etc/zabbix/zabbix_agent2.d',
      'pidfile'               => '/var/run/zabbix/zabbix_agentd2.pid',
      'servicename'           => 'zabbix-agent2',
      'zabbix_package_agent'  => 'zabbix-agent2',
      'binary_location'       => '/usr/sbin/zabbix_agent2',
```

I actually not test with windows kernel.

#### This Pull Request (PR) fixes the following issues
Fixes #692

PR #929, opened 2024-03-05
Bootstrap agent2 support

### LINKED ISSUE #692: Add Support for Zabbix Agent 2

Valantin:
This pull Request was a rebase of work in #914 and other

* Add param binary_location for specify zabbix agent binary (only for systemd)
* Remove some parameters in zabbix agent configuration if it's zabbix agent2 package * LogRemoteCommands * StartAgents * MaxLinesPerSecond * AllowRoot * User * LoadModulePath
* Fix EnableRemoteCommands still use on zabbix agentd > 5 but remove on agent2

Conflicts:
	manifests/agent.pp
	spec/classes/agent_spec.rb
	templates/zabbix_agentd.conf.erb

Fixes #692


manifests/startup.pp
```diff
@@ -11,6 +11,7 @@
 # @param service_type Systemd service type
 # @param manage_database When true, it will configure the database and execute the sql scripts.
 # @param service_name Name of the service. Defaults to the resource name
+# @param binary_location This params is use for define a specific binary location. This is actually only available for zabbix agent and systemd
```
bastelfreak, line 14: do we need the whole startup.pp for the new agent? doesn't it ship with a ready to use unit file?

manifests/init.pp
```diff
@@ -200,18 +200,31 @@
...
+#
+# @example Using Zabbix Agent 2
+#   class { 'zabbix::agent':
+#     package_name          => 'zabbix-agent2',
+#     agent_configfile_path => '/etc/zabbix/zabbix_agent2.d',
+#     pidfile               => '/var/run/zabbix/zabbix_agentd2.pid',
+#     servicename           => 'zabbix-agent2',
+#     zabbix_package_agent  => 'zabbix-agent2',
+#     binary_location       => '/usr/sbin/zabbix_agent2'
+#   }
```
h-haaks, line 226: Should this example be moved to the agent class?

manifests/startup.pp
Valantin: Ok, I've checked the deb package and both zabbix-agent and zabbix-agent2 supply a system-unit
So, probably is needed to remove the unit dropin built in the startup class or add a conditional parameter to use it if true and default to false

What do you think is better?

bastelfreak: I think we should not touch the existing setup for the old agent. That would become complicated. But if the new agent ships a unit, we should try to use it.

bastelfreak:
Please also take a look at https://github.com/voxpupuli/puppet-zabbix/tree/master/spec/acceptance and add an acceptance test for the new agent.

manifests/init.pp
Valantin: moved

manifests/startup.pp
Valantin: Ok, I't now use the code shipped with the module adding just the way to change the path, so, probably I can add a condition using agent2 flag to skip unit creation

manifests/agent.pp
```diff
@@ -134,6 +134,16 @@
 #    monitored_by_proxy => 'my_proxy_host',
 #    server             => '192.168.1.1',
 #  }
+#
+# @example Using Zabbix Agent 2
+#   class { 'zabbix::agent':
+#     agent_configfile_path => '/etc/zabbix/zabbix_agent2.conf',
+#     include_dir           => '/etc/zabbix/zabbix_agent2.d',
+#     pidfile               => '/var/run/zabbix/zabbix_agentd2.pid',
```
bastelfreak, line 142: is it actually required to pass the pidfile?

Valantin: Yes, default PID was on /tmp but for example on ubuntu22.04 unit file use different file

manifests/startup.pp
Valantin: it seems to fail on non debian os, check it deeply

Valantin: So, different OS use different location for pid in the unit file
what you think is the best way to approch this problem?

making custom unit mask it

Phil-Friderici:
I'm very happy to see all the activities here now.
If you need help, I can try to work on the unit tests. Let me know.

spec/acceptance/agent_spec.rb
```diff
@@ -87,4 +98,127 @@ class { 'zabbix::agent':
...
+              zabbix_package_state  => 'latest',
+              listenip              => '127.0.0.1',
+              zabbix_version        => '#{version}',
+            }
+          EOS
+
+          prepare_host
+
+          apply_manifest(pp, catch_failures: true)
+          apply_manifest(pp, catch_changes: true)
```
bastelfreak, line 215: Can you please use our shared example please? https://github.com/voxpupuli/voxpupuli-acceptance?tab=readme-ov-file#an-idempotent-resource

Valantin: I'm not able to run function prepare_host in it_behaves_like like I do in the standard it 
I've try with "prepare_host" before the it_behaves_like, with a "before do" statement but it apply the code every time

continue trying, but if you have the solution please share it

spec/acceptance/agent_spec.rb
```diff
@@ -44,6 +46,9 @@ class { 'zabbix::agent':
             zabbix_version       => '#{version}',
           }
         EOS
+
+        prepare_host
```
bastelfreak, line 50: is that call needed here?

spec/acceptance/agent_spec.rb
```diff
@@ -63,6 +68,9 @@ class { 'zabbix::agent':
...
```
bastelfreak, line 72: same here, is that actually needed? It worked before

spec/acceptance/agent_spec.rb
```diff
@@ -77,6 +85,9 @@ class { 'zabbix::agent':
               zabbix_version       => '#{version}',
             }
           EOS
+
+          prepare_host
```
bastelfreak, line 89: same here

spec/support/acceptance/prepare_host.rb
```diff
@@ -5,21 +5,28 @@ def prepare_host
...
+      'RedHat' => ['zabbix-server-pgsql', 'zabbix-server-pgsql-scl', 'zabbix-web', 'zabbix-web-pgsql', 'zabbix-web-pgsql-scl', 'zabbix-frontend-php', 'zabbix-sql-scripts', 'zabbix-agent', 'zabbix-agent2'],
+      'Debian' => ['zabbix-server-pgsql', 'zabbix-web-pgsql', 'zabbix-frontend-php', 'zabbix-sql-scripts', 'zabbix-agent', 'zabbix-agent2'],
+      default  => ['zabbix-agent', 'zabbix-agent2'],
     }
+
+    $pkg_ensure = $facts['os']['family'] ? {
+      'RedHat' => purged,
+      'Debian' => purged,
+      default  => absent,
+    }
```
bastelfreak, line 26: Is that block required? Which OS doesn't support `purged`?

Valantin: Archlinux not support purged, I don't check the other, you prefer to use purged as default and set only on archlinux absent?

spec/acceptance/agent_spec.rb
Valantin: Yes, but I'm work on the other indempotent test to use shared example, if it was merged I can rebase this code

spec/support/acceptance/prepare_host.rb
bastelfreak: Oh this is indeed a bug within Puppet

bastelfreak: https://github.com/puppetlabs/puppet/pull/9287

Valantin: I've restore purged as default, it fail on archlinux before the PR was merged and released

Valantin: While waiting for the release I set purged as default and absent for Archlinux

PR #935, opened 2024-03-17
Test docker_podman support in beaker

bastelfreak:


spec/spec_helper_acceptance.rb
```diff
@@ -6,7 +6,7 @@
   # In Puppet 7 the locale ends up being C.UTF-8 if it isn't passed.
   # This locale doesn't exist in EL7 and won't be supported either.
   # At least PostgreSQL runs into this.
-  ENV['LANG'] = 'en_US.UTF-8' if host['hypervisor'] == 'docker' && host['platform'] == 'el-7-x86_64'
+  ENV['LANG'] = 'en_US.UTF-8' if host['hypervisor'] =~ %r{docker} && host['platform'] == 'el-7-x86_64'
```
evgeni, line 9: should this be `=~ container`? not sure what exactly is the value in `hypervisor` here.

bastelfreak: I also don't know. I didn't pay too much attention because I want to merge https://github.com/voxpupuli/puppet-zabbix/pull/970 first, and that will remove the whole line anyways.

evgeni: fair.

it still would be good to know as I think we use similar blocks in other modules too

PR #947, opened 2024-05-17
Support multiple values in TLSAccept

wolfaba:
Option TLSAccept in zabbix_agentd.conf accepts list of values separated with comma. But the module define tlsaccept as Enum of those three values. The string as comma separated list of values is not allowed anymore. Therefore the tlsaccept should be array of that enum and in the template the array should be joined with comma.

BTW: the same option TLSAccept is in zabbix_proxy.conf, but in the module the tlsaccept in proxy.pp is still simple string value.

Thank you.

### DIFF
--- a/templates/zabbix_agentd.conf.erb
+++ b/templates/zabbix_agentd.conf.erb
@@ -310,7 +310,7 @@ LoadModulePath=<%= @loadmodulepath %>
 # Mandatory: yes, if TLS certificate or PSK parameters are defined (even for 'unencrypted' connection)
 # Default:
 # TLSAccept=unencrypted
-<% if @tlsaccept %>TLSAccept=<%= @tlsaccept %><% end %>
+<% if @tlsaccept %>TLSAccept=<%= [@tlsaccept].flatten.join(',') %><% end %>
 
 ### Option: TLSCAFile
 #	Full pathname of a file containing the top-level CA(s) certificates for
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1860,7 +1860,7 @@ Default value: `$zabbix::params::agent_timeout`
 
 ##### <a name="-zabbix--agent--tlsaccept"></a>`tlsaccept`
 
-Data type: `Optional[Enum['unencrypted','psk','cert']]`
+Data type: `Optional[Variant[Array[Enum['unencrypted','psk','cert']],Enum['unencrypted','psk','cert']]]`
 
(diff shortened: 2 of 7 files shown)


wolfaba:
I am sorry for so many commits. I am neither ruby nor puppet developer :disappointed: If you understand, what I mean, could you please fix the "if"? I need join array values or just insert single string (Enum) value. Thank you.

templates/zabbix_agentd.conf.erb
```diff
@@ -310,7 +310,7 @@ LoadModulePath=<%= @loadmodulepath %>
 # Mandatory: yes, if TLS certificate or PSK parameters are defined (even for 'unencrypted' connection)
 # Default:
 # TLSAccept=unencrypted
-<% if @tlsaccept %>TLSAccept=<%= @tlsaccept %><% end %>
+<% if @tlsaccept and ( @tlsaccept.instance_of?(Array) or @tlsaccept.instance_of?(String) ) %>TLSAccept=<% if @tlsaccept.instance_of?(Array) %><%= scope.call_function('join', [@tlsaccept, ',']) %><% elsif @tlsaccept.instance_of?(String) %><%= @tlsaccept %><% end %><% end %>
```
smortex, line 313: Something like this (not tested)?

```suggestion
<% if @tlsaccept %>TLSAccept=<%= [@tlsaccept].flatten.join(',') %><% end %>
```

wolfaba: @smortex of course it works, thank you very much, it's simple and nice :+1:

Valantin:
ok, now I can you add spec test for the classes?

for agent exist
`it { is_expected.to contain_file('/etc/zabbix/zabbix_agentd.conf').with_content %r{^TLSAccept=cert$} }`
You can try to add two test, using tlsaccept => ['cert'] and one using a two element array.
the first one is for check it work in the some way with string and array with one element and the second for check it costruct correctly the string with an array of two or more.
For proxy not exist the spec, probably you can reuse the one from the agent

```
 context 'configuration file with full options' do
    if facts[:kernel] == 'Linux'
      let :params { tlsaccept: ['cert'] }
      it { is_expected.to contain_file('/etc/zabbix/zabbix_agentd.conf').with_content %r{^TLSAccept=cert$} }
    end
 end
```

Valantin (approved):

Valantin:
@wolfaba can you squash in one commit with good message?

wolfaba:
@Valantin is it better now? I hope I haven't broke it :grimacing:

PR #952, opened 2024-06-17
Add zabbix 7.0 support in CI

Valantin:
Add zabbix 7.0 to CI

Zabbix has released stable lts 7.0 not tested in CI

zabbixapi gem not support zabbix 7 so it fail if you use the resource managed by it https://github.com/voxpupuli/puppet-zabbix/blob/051f24c0ec94ff1a45e75e3ad58c42eb30b41fa1/manifests/zabbixapi.pp#L15

### DIFF
--- a/spec/acceptance/zabbix_host_spec.rb
+++ b/spec/acceptance/zabbix_host_spec.rb
@@ -6,6 +6,9 @@ require 'serverspec_type_zabbixapi'
 # rubocop:disable RSpec/LetBeforeExamples
 describe 'zabbix_host type' do
   supported_server_versions(default[:platform]).each do |zabbix_version|
+    # Zabbix 7.0 removed the deprecated params 'user' in favor to 'username'
+    next if zabbix_version >= '7.0'
+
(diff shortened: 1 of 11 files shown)


spec/classes/database_postgresql_spec.rb
```diff
@@ -43,7 +43,7 @@
                end
 
         sql_server = case zabbix_version
-                     when '6.0'
+                     when '6.0', '7.0'
```
bastelfreak, line 46: does it make more sense to check for >= 6 ?

spec/acceptance/zabbix_host_spec.rb
```diff
@@ -6,6 +6,8 @@
 # rubocop:disable RSpec/LetBeforeExamples
 describe 'zabbix_host type' do
   supported_server_versions(default[:platform]).each do |zabbix_version|
+    next if zabbix_version >= '7.0'
```
bastelfreak, line 9: can you add a comment why we skip this?

bastelfreak (approved):

PR #955, opened 2024-06-17
Add/ubuntu24

Valantin:
Add Ubuntu 24.04 to supported OS

#### This Pull Request (PR) include
#952


spec/spec_helper_methods.rb
```diff
@@ -13,7 +13,7 @@ def baseline_os_hash
       },
       {
         'operatingsystem' => 'Ubuntu',
-        'operatingsystemrelease' => %w[20.04 22.04]
+        'operatingsystemrelease' => %w[20.04 22.04 24.04]
```
bastelfreak, line 16: I think in another PR we should remove the whole hash here. It doesn't make sense to duplicate the information from metadata.json. We did that in the past because we only had limited CI power.

Valantin: yes but we need to remove not supported os from metadata.json
I'm waiting centos7 EOL to think to release v11

bastelfreak: FYI: With https://github.com/voxpupuli/puppet-zabbix/pull/967 the unit tests will run much faster, so we don't need to worry about the CI runtime for unit tests.

bastelfreak:
@baron1405 This PR needs to be rebased before we can merge it. If you are interested in it, you can checkout the branch, rebase it against our latest master branch and submit a new PR.

baron1405:
@bastelfreak Seems like I would be the least qualified to perform the merge. I have a highly customized puppet environment so developing for the forge environment is something a bit new to me. If you provide instructions on what I should do, I can see if I feel comfortable doing it.

bastelfreak:
I don't think you need much Puppet knowledge here. The PR just adds ubuntu 24.04 to two files and those merge conflicts need to be resolved. If you have any git questions you can come to our IRC or slack channel (https://voxpupuli.org/connect/) and we can figure it out together.

PR #959, opened 2024-06-25
feat: allow uninstalling the Zabbix agent

coreone:
Add the ability to remove the Zabbix agent from a system using the `ensure` parameter.

feat: add tests for new "ensure" parameter in zabbix::agent


Valantin:
Hi @coreone have you tried with `zabbix::agent::zabbix_package_state: purged`

coreone:
I have done that, and it removes the package(s), but it doesn't remove the config file. Normally, I would just leave the config file there, but when you apply the agent class with the following parameters, it causes a runtime error when applying the catalog because the `/etc/zabbix` directory doesn't exist (because the package doesn't exist, and that's what creates the directory) and it still tries to write the config file.

```puppet
manage_repo => false,
service_enable => false,
service_ensure => 'stopped',
zabbix_package_state => 'absent',
```

PR #969, opened 2024-06-30
Run acceptance tests on Ubuntu 20.04 M runners

bastelfreak:
Those runners contain 4 cores / 16G ram. The default S runners only have 2 cores.


TheMeier (approved):

d1nuc0m:
@bastelfreak @TheMeier is this still relevant?

PR #970, opened 2024-07-01
Remove EL7 support

Valantin:

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -21,9 +21,13 @@ class zabbix::repo (
     case $facts['os']['family'] {
       'RedHat': {
         $majorrelease = $facts['os']['release']['major']
-        if (versioncmp(fact('os.release.major'), '7') >= 0 and $zabbix_version == '7.0') {
+        if versioncmp($zabbix_version, '7.0') >= 0 {
           $gpgkey_zabbix = 'https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-B5333005'
-          $gpgkey_nonsupported = 'https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-B5333005'
+          if versioncmp(fact('os.release.major'), '9') >= 0 {
+            $gpgkey_nonsupported = 'https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-08EFA7DD'
+          } else {
+            $gpgkey_nonsupported = 'https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-B5333005'
+          }
         } elsif versioncmp(fact('os.release.major'), '9') >= 0 {
           $gpgkey_zabbix = 'https://repo.zabbix.com/RPM-GPG-KEY-ZABBIX-08EFA7DD'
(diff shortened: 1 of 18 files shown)


manifests/repo.pp
```diff
@@ -21,7 +21,7 @@
     case $facts['os']['family'] {
       'RedHat': {
         $majorrelease = $facts['os']['release']['major']
-        if (versioncmp(fact('os.release.major'), '7') >= 0 and $zabbix_version == '7.0') {
+        if (versioncmp(fact('os.release.major'), '8') >= 0 and $zabbix_version == '7.0') {
```
bastelfreak, line 24: if we don't support EL7, do we need the check for `versioncmp(fact('os.release.major'), '8'`?

spec/classes/repo_spec.rb
```diff
@@ -84,52 +84,49 @@
         end
 
         case facts[:os]['release']['major']
-        when '7'
-          context 'on RedHat 7 and Zabbix 5.0' do
+        when '8', '9'
```
bastelfreak, line 87: the whole case statement can go away?

manifests/repo.pp
```diff
@@ -21,7 +21,7 @@
     case $facts['os']['family'] {
       'RedHat': {
         $majorrelease = $facts['os']['release']['major']
-        if (versioncmp(fact('os.release.major'), '7') >= 0 and $zabbix_version == '7.0') {
+        if versioncmp($zabbix_version, '7.0') >= 0 {
```
bastelfreak, line 24: the whole block still doesn't correct to me. first we check on the zabbix version, then on the OS version and then an else block?

bastelfreak:
@Valantin can you please rebase against our master branch?

Valantin:
Yes ASAP, I'm without computer

Valantin:
@bastelfreak sorry for the delay

bastelfreak:
@Valantin thanks for the rebase, can you please take a look at the failing tests?

Valantin:
Yes, I think is something related to the repository key, hope to check it in the afternoon

Valantin:
EL9 issue seems to be related to postgres, I've tried alma9 with vagrant_libvirt and it fail because not found postgres13 module
set default to 15 it fail for locale
on the vm
```
[root@almalinux9-64-1 ~]# localectl status
System Locale: LANG=C.UTF-8
    VC Keymap: us
   X11 Layout: us
```

next week try to fix

steven-letsengage:
Just as a FYI, it's not uncommon for enterprise customers to purchased 4 years extend support, RHEL 7 will still be around until 2028.
https://www.redhat.com/en/blog/announcing-4-years-extended-life-cycle-support-els-red-hat-enterprise-linux-7

bastelfreak:
Hi,
In https://voxpupuli.org/blog/2024/05/08/CentOS-7-CentOS-Stream-8-deprecation/ we agreed on deprecating RHEL 7. The extended support lifetime is very limited and we have no way to test it in our CI, also people had enough time to migrate to newer operating systems. RHEL 8 was released in 2019.

Valantin:
@bastelfreak the problem with the LANG was just on my vagrant test environment.

fixing the version of Postgres module now all test pass

PR #977, opened 2024-10-16
Add Authorization header if needed to Apache vhost

DJClean:
On Zabbix version 6.4 and newer there has been a change in the way
authentication is done. Using a bearer token would lead to issues with
authenticating using the API as the header did not seem the get passed to
the application.

This seems to have been known since the release of 6.4 and a bug report
is still open. But the bug report also already contained a
workaround/fix (https://support.zabbix.com/browse/ZBX-22952).

n/a


DJClean:
Seems this is resolved in the latest version and the PR can be closed

PR #978, opened 2024-10-24
Enable 4 new 7+ server parameters

trenta:
This PR enables 4 new server parameters for version >= 7
StartAgentPollers
StartHTTPAgentPollers
StartSNMPPollers
MaxConcurrentChecksPerPoller


fraenki:
```REFERENCE.md is outdated; to regenerate: bundle exec rake strings:generate:reference```

@trenta Please fix the test failure(s) and force-push to trigger the test suite :)

**Comment by trenta** [comment]
Is that what you were after?

**Comment by fraenki** [comment]
Looks better now, but almost all tests for Zabbix 7.0 are failing. Looking at [this example](https://github.com/voxpupuli/puppet-zabbix/actions/runs/12366185644/job/34512479550?pr=978), the reason is probably always the same:

```
  deploys a zabbix 7.0 server
    behaves like an idempotent resource
...
  Notice: /Stage[main]/Zabbix::Repo/Yumrepo[zabbix]/baseurl: baseurl changed 'https://repo.zabbix.com/zabbix/6.0/rhel/8/$basearch/' to 'https://repo.zabbix.com/zabbix/7.0/rhel/8/$basearch/'
...

  Error: /Stage[main]/Zabbix::Server/Service[zabbix-server]: Failed to call refresh: Systemd restart for zabbix-server failed!
  journalctl log for zabbix-server:
  -- Logs begin at Tue 2024-12-17 04:10:07 UTC, end at Tue 2024-12-17 04:12:37 UTC. --
  Dec 17 04:12:28 rocky8-64-puppet8.example.com systemd[1]: Starting Zabbix Server...
  Dec 17 04:12:28 rocky8-64-puppet8.example.com zabbix_server[3919]: Starting Zabbix Server. Zabbix 6.0.36 (revision 6bcc7a62c0e).
```

Although the test is expected to setup Zabbix Server 7.0, it ends up starting version 6.0.36. I assume that this has always been the case, but this PR added options that are only available in version 7 and up, making this issue visible.

Could you probably have a look at the tests?

trenta:
I'm happy to participate in a fix but not sure it should be part of this pr. I haven't had any experiences with test coding. I had a look though and to me it looks like server version 6.0 is the only version for testing. If we are to add a very 7 as well would we need to make sure the correct tests are associated with correct version? This is where I'm not sure and don't wan to break things.

PR #973, opened 2024-07-09
Use Sensitive for Secrets

cocker-cc:
#### Pull Request (PR) description
To not reveal Secrets, accept Datatype `Sensitive`.  Render Templates as sensitive Content, if Secrets were given as `Sensitive`.

#### This Pull Request (PR) fixes the following issues
Fixes #950

PR #979, opened 2024-10-25
Prepare Release 11.0.0

jstraw:
Release 10.0.0 to:
*  manage startup scripts
* Remove EOL and unsupported OSes
* Accept multiple TLSAccept options

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,41 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
-## [v10.0.0](https://github.com/voxpupuli/puppet-zabbix/tree/v10.0.0) (2024-03-14)
+## [v11.0.0](https://github.com/voxpupuli/puppet-zabbix/tree/v11.0.0) (2024-10-24)
+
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):

teluq-pbrideau:
I’m quite frustrated to see a new release done without my pending changes.

I’ve got a PR that is stale (#857) for two years now, I’ve rebased multiple time to make sure it is ready to get merged, but this never get done.
There is another PR with the same issue by @cocker-cc, but with a different approach (not as heavy handed as mine), (#973) that is also not moving forward.

DJClean:
Are there any plans for this to do this release and perhaps also include several PRs that are waiting to solve small issues or add params from the latest version?

FunPat:
Any chance the airlocked environment change I pushed through can make the next puppet forge release? All good if not. Using commit based ref at the moment in the puppet file so its not a huge drama but would love to use the forge based releases in the puppetfile if possible as we use the r10k dependency checker in the devtools provided by the voxpupuli to keep modules up to date and it doesn't handle non forge refs all that well.
"Unable to determine ref type for puppet-zabbix"

PR has been accepted and merged, just not baselined for this release.
https://github.com/voxpupuli/puppet-zabbix/pull/987

Thanks 👍

TheMeier:
superseeded by #996

PR #987, opened 2025-02-18
Add alternative location for GPG keys for the zabbix and unsupported

FunPat:
Added two variables to allow users of this module to set an alternative resource location for the GPG keys, when operating in an airlocked environment where servers cannot talk out to repo.zabbix.com directly DNF/YUM will fail to check the GPG key check if the repositories are managed by this module. This will allow an internal GPG key to be referenced so that the puppet-zabbix module can be used on servers that cannot have direct internet access, this is especially important from an agent perspective.

Variables introduced:
- repo_gpg_key_location
- unsupported_repo_gpg_key_location

Two checks have been introdcued under the repo.spec for RedHat to test that when these have been set that the GPG key has the alternative location set. Used https://example.com/bar to represent if it is a different location to where the repository is, ie repository location being https://example.com/foo.

No current issues relating to this that I'm aware of.


FunPat:
Ready for review when someone is available, thanks in advance. Any questions, concerns please let me know.
Love what the voxpupli community do! I hope this is a valuable addition to the module, would certainly assist in the environment I work in :)

evgeni (approved):

PR #988, opened 2025-02-19
Add SMSDevices in zabbix server 7+

### LINKED ISSUE #985: SMSDevices required in zabbix_server.conf
Since Zabbix 7 you're required to define SMSDevices in the zabbix_server.conf
a simple config option to add those would be good

From the zabbix docu:

```
SMSDevices
A list of comma-delimited modem files allowed to use Zabbix server.
SMS sending is not possible if this parameter is not set.

Example:

 SMSDevices=/dev/ttyUSB0,/dev/ttyUSB1
```

Ph0B1uS:
Adds SMSDevices to zabbix_server.conf if defined for zabbix_version 7+

Fixes #985

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -116,6 +116,7 @@
 # @param starttimers Number of pre-forked instances of timers.
 # @param javagateway IP address (or hostname) of zabbix java gateway.
 # @param javagatewayport Port that zabbix java gateway listens on.
+# @param smsdevices Which devices to use for sms texting
 # @param startjavapollers Number of pre-forked instances of java pollers.
 # @param startlldprocessors Number of pre-forked instances of low-level discovery (LLD) workers.
 # @param startvmwarecollectors Number of pre-forked vmware collector instances.
@@ -304,6 +305,7 @@ class zabbix (
   $cachesize                                                                  = $zabbix::params::server_cachesize,
   $cacheupdatefrequency                                                       = $zabbix::params::server_cacheupdatefrequency,
   $startdbsyncers                                                             = $zabbix::params::server_startdbsyncers,
+  Array[String[1]] $smsdevices                                                = $zabbix::params::server_smsdevices,
   $historycachesize                                                           = $zabbix::params::server_historycachesize,
   Zabbix::Historyics $historyindexcachesize                                   = $zabbix::params::server_historyindexcachesize,
   $trendcachesize                                                             = $zabbix::params::server_trendcachesize,
@@ -423,6 +425,7 @@ class zabbix (
     database_port             => $database_port,
     database_tlsconnect       => $database_tlsconnect,
     database_tlscafile        => $database_tlscafile,
+    smsdevices                => $smsdevices,
     startpollers              => $startpollers,
     startipmipollers          => $startipmipollers,
     startpollersunreachable   => $startpollersunreachable,
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -240,6 +240,7 @@ class zabbix::params {
   $server_pacemaker_resource                = undef
   $server_snmptrapperfile                   = '/tmp/zabbix_traps.tmp'
   $server_sourceip                          = undef
+  $server_smsdevices                        = []
   $server_sshkeylocation                    = undef
   $server_sslcertlocation                   = '/usr/lib/zabbix/ssl/certs'
   $server_sslkeylocation                    = '/usr/lib/zabbix/ssl/keys'
--- a/spec/classes/server_spec.rb
+++ b/spec/classes/server_spec.rb
@@ -367,6 +367,28 @@ describe 'zabbix::server' do
         it { is_expected.to contain_file('/etc/zabbix/zabbix_server.conf').without_content %r{^StartODBCPollers=1} }
       end
 
+      context 'with zabbix_server.conf and version 7.0' do
+        let :params do
+          {
+            smsdevices: ['/dev/ttyS0'],
+            zabbix_version: '7.0'
+          }
+        end
+
+        it { is_expected.to contain_file('/etc/zabbix/zabbix_server.conf').with_content %r{^SMSDevices=/dev/ttyS0} }
+      end
+
+      context 'with zabbix_server.conf and version 7.2' do
+        let :params do
+          {
+            smsdevices: ['/dev/ttyS0'],
+            zabbix_version: '7.2'
+          }
+        end
+
+        it { is_expected.to contain_file('/etc/zabbix/zabbix_server.conf').with_content %r{^SMSDevices=/dev/ttyS0} }
+      end
+
       context 'with zabbix_server.conf and logtype declared' do
         describe 'as system' do
           let :params do
--- a/templates/zabbix_server.conf.erb
+++ b/templates/zabbix_server.conf.erb
@@ -962,8 +962,10 @@ LoadModulePath=<%= @loadmodulepath %>
 # Mandatory: no
 # Default:
 # SMSDevices=
+<% if @smsdevices.length > 0 -%>
+SMSDevices=<%= @smsdevices.join(',') -%><% end -%>
+<% end -%>
 
-<% end %>
 <% if @zabbix_version.to_f >= 5.4 %>
 ####### For advanced users - TCP-related fine-tuning parameters #######
 
--- a/.msync.yml
+++ b/.msync.yml
@@ -2,4 +2,4 @@
 # Managed by modulesync - DO NOT EDIT
 # https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
 
-modulesync_config_version: '9.4.0'
+modulesync_config_version: '9.5.0'
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ group :test do
   gem 'voxpupuli-test', '~> 9.0',   :require => false
   gem 'coveralls',                  :require => false
   gem 'simplecov-console',          :require => false
-  gem 'puppet_metadata', '~> 4.0',  :require => false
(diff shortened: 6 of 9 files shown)


templates/zabbix_server.conf.erb
```diff
@@ -911,6 +911,7 @@ LoadModulePath=<%= @loadmodulepath %>
 <% if @webserviceurl %>WebServiceURL=<%= @webserviceurl -%><% end %>
 <% end %>
 
+<%# FIXME: vi får v 6.0 %>
```
evgeni, line 914: ? :)

manifests/init.pp
```diff
@@ -304,6 +305,7 @@
   $cachesize                                                                  = $zabbix::params::server_cachesize,
   $cacheupdatefrequency                                                       = $zabbix::params::server_cacheupdatefrequency,
   $startdbsyncers                                                             = $zabbix::params::server_startdbsyncers,
+  Optional[String] $smsdevices                                                = $zabbix::params::server_smsdevices,
```
evgeni, line 308: I wonder if this should be `Optional[Array[String]]` so people can use `['/dev/ttyUSB0', '/dev/ttyUSB1']` and the template joins them with a comma -- feels more natural to pass an array?

Also, @bastelfreak will tell you this should be at lest `String[1]` as an empty string won't be a good device :)

templates/zabbix_server.conf.erb
Ph0B1uS: I'll remove this, thanks for pointing it out.

manifests/init.pp
Ph0B1uS: I'm unsure about this as the config file takes a comma delimited string.
The template should not output anything if $smsdevices is empty unless the erp if syntax is different to that of epp.

templates/zabbix_server.conf.erb
Ph0B1uS: Says in Swedish: "we're getting 6.0" in case anyone is curious :)

manifests/init.pp
evgeni: I think in the template you still could do `<% if @smsdevices -%>SMSDevices=<%= @smsdevices.join(',') %><% end %>`

Wouldn't block on this, but curious what the others think.

ekohl: I'd prefer `Array[String[1]]` (default value `[]`) and then in the template use:
```erb
<% if @smsdevices.any? -%>
SMSDevices=<%= @smsdevices.join(',') %>
<% end %>

This way it's not shown if there is an empty array (which is the default value), but still shows all values.

Comment (Ph0B1uS) [reply manifests/init.pp]: Think I managed to sort it out now, tests are failing but that seems to be due to the forge not responding and there's not a lot I can do about that :)

**Comment by Valantin** [comment]
Hi, can you squash the commit and use just the verified account?

File: templates/zabbix_server.conf.erb [hunk 3]
```diff
@@ -962,8 +962,10 @@ LoadModulePath=<%= @loadmodulepath %>
 # Mandatory: no
 # Default:
 # SMSDevices=
+<% if @smsdevices -%>
+SMSDevices=<%= @smsdevices %><% end -%>
```
ekohl, line 966: Shouldn't this join the entries with commas? Also, if it's an empty array, will this work well?

Ph0B1uS: it should not output anything if the array is empty unless there's something that i've missed

Ph0B1uS:
> Hi, can you squash the commit and use just the verified account?

I did manage to change to the verified account but it seems the squash was unsuccessful as i can still see the old commits in the repo.
I tried `git rebase -i HEAD~16`

Valantin:
@Ph0B1uS have you replace the pick to the squash?

Ph0B1uS:
> @Ph0B1uS have you replace the pick to the squash?

Yup, i did that and changed the commit message too, then push -f but it only added to the commit count I'm afraid.

Ph0B1uS:
Alright to merge or are there any suggestions and/or corrections that needs to be implemented?

Valantin:
Sorry but it can't be merged without the required changes, you can find here https://voxpupuli.org/docs/reviewing_pr/ the list of things to check before approve a PR

Ph0B1uS:
> Sorry but it can't be merged without the required changes, you can find here https://voxpupuli.org/docs/reviewing_pr/ the list of things to check before approve a PR

The only two points I can find in that document that needs to be rectified is: 
1) the label and correct me if I'm wrong but it would appear as if I cannot do that as in am not allowed to.
2) the unfortunate commits from a non approved account.

How about I close this PR, open a new one to rectify the 2nd issue. Would that work or is there something else I've missed?

manifests/init.pp
```diff
@@ -304,6 +305,7 @@
   $cachesize                                                                  = $zabbix::params::server_cachesize,
   $cacheupdatefrequency                                                       = $zabbix::params::server_cacheupdatefrequency,
   $startdbsyncers                                                             = $zabbix::params::server_startdbsyncers,
+  Optional[Array[String]] $smsdevices                                         = $zabbix::params::server_smsdevices,
```
bastelfreak, line 308: The datatype is inconsistent:
```suggestion
  Optional[Array[String[1]]] $smsdevices                                      = $zabbix::params::server_smsdevices,
```

Valantin:
try to use "git rebase -i HEAD~3" replace "pick" with "edit" for the first line "ADD..." and replace "pick" with "squash" for all the other line, than save, "git commit --amend --author "Ph0B1uS <bjorn.thoor@gmail.com>" --no-edit" than "git rebase --continue"

If all is ok "git log" show you one commit than you can push it

manifests/init.pp
Ph0B1uS: sorted, thanks for pointing it out

Ph0B1uS:
Reopening to rerun tests

Ph0B1uS:
Reopening to rerun tests, hopefully they work a bit better this time

Ph0B1uS:
The failing tests has nothing to do with my code as they stem from a file which I have not touched, now what? :)

templates/zabbix_server.conf.erb
```diff
@@ -962,8 +962,10 @@ LoadModulePath=<%= @loadmodulepath %>
 # Mandatory: no
 # Default:
 # SMSDevices=
+<% if @smsdevices -%>
```
bastelfreak, line 965: When `smsdevices` is an empty array, this evaluates to true? And than we get `SMSDevices=` in the config?

manifests/params.pp
```diff
@@ -240,6 +240,7 @@
   $server_pacemaker_resource                = undef
   $server_snmptrapperfile                   = '/tmp/zabbix_traps.tmp'
   $server_sourceip                          = undef
+  $server_smsdevices                        = []
```
Valantin, line 243: I think is better to set it to undef

Valantin (changes requested):
Add test for the new parameter if it is on zabbix >= 7.0

manifests/params.pp
bastelfreak: If you can provide multiple values, an empty is usually preferred. see https://voxpupuli.org/docs/reviewing_pr/
> Are there new parameters with datatype Hash or Array? If possible, they should default to empty Hash or Array instead of undef. You can also enforce the datastructure like Array[String[1]] (that’s recommended if possible)

The check in the template just needs to check if it's an empty array.

templates/zabbix_server.conf.erb
Ph0B1uS: Updated the line to <% if @smsdevices.length > 0 -%>

manifests/params.pp
Ph0B1uS: I believe this should take care of checking the length of the array:

`<% if @smsdevices.length > 0 -%>`

https://github.com/Ph0B1uS/puppet-zabbix/blob/6e76728cd7205a8dc0e46e7a39a81a70b638fda3/templates/zabbix_server.conf.erb#L965

Ph0B1uS: As for the tests, I do not know how to do that I'm afraid.

Ph0B1uS: I believe the 7.0 parameter tests works now and I also added 7.2 dittos

spec/classes/server_spec.rb
```diff
@@ -367,6 +367,34 @@
...
+            smsdevices: ['/dev/ttyS0'],
+            zabbix_version: '7.0'
+          }
+        end
+
+        it { is_expected.to contain_file('/etc/zabbix/zabbix_server.conf').with_content %r{^SMSDevices=/dev/ttyS0} }
+      end
+
+      context 'with zabbix_server.conf and version 7.2' do
+        next if facts[:os]['name'] == 'Debian' && facts[:os]['release']['major'].to_i <= 12
```
bastelfreak, line 385: Please see my comment in https://github.com/voxpupuli/puppet-zabbix/pull/991#discussion_r1987044848 . the OS check shouldn't be in the tests. The zabbix version check needs to be in the template. otherwise we will get a broken config on zabbix 6.x and older and the tests will mask that.

jstraw (approved):

manifests/init.pp
```diff
@@ -304,6 +305,7 @@
   $cachesize                                                                  = $zabbix::params::server_cachesize,
   $cacheupdatefrequency                                                       = $zabbix::params::server_cacheupdatefrequency,
   $startdbsyncers                                                             = $zabbix::params::server_startdbsyncers,
+  Optional[Array[String[1]]] $smsdevices                                      = $zabbix::params::server_smsdevices,
```
bastelfreak, line 308:
```suggestion
  Array[String[1]] $smsdevices                                                = $zabbix::params::server_smsdevices,
```

PR #1002, opened 2025-06-10
fix: SELinux httpd LDAP boolean

d1nuc0m:

Set SELinux boolean to allow for LDAP authentication

n/a


d1nuc0m:
Thank you, could you also have a look at the pipeline? I don't know why it's failing in this specific case only

FunPat:
@d1nuc0m I have seen issues personally like that with Windows line endings (^M) being present in the edited files from the source control check out, although generally more tests fail. It might just be a matter of retriggering the tests, ie a minor change then change it back to retrigger (not aware of a way to trigger these manually outside of that or reopening the pull request)

Regarding the line endings if you want to check: If you are on windows, you can use WSL (Windows Subsystem for Linux) with "cat -v <file>" to see the line endings, ^M and then you can use dos2unix on that file if that is the case to convert the line endings. Otherwise some IDEs might be able to show you that. Otherwise you can use cat directly on linux to check.

d1nuc0m:
@FunPat thank you, I'm working on EL, also checked for windows line endings and there seem to be none.
If you check the actions the checks have already been ran a couple times by @bastelfreak with the same result; I've also tried to recreate it in local but here it works

d1nuc0m:
@FunPat update - I've been able to reproduce it locally with `BEAKER_PUPPET_COLLECTION=puppet8 BEAKER_destroy=no BEAKER_SETFILE="almalinux9-64" BEAKER_HYPERVISOR=docker BEAKER_FACTER_zabbix_version="6.0" SPEC="spec/acceptance/zabbix_host_spec.rb"  bundle exec rake beaker`  and I'm trying to debug it, if you'd like to participate in the discussion you may head on IRC to Libera.Chat #voxpupuli

d1nuc0m:
@bastelfreak @FunPat finally found the issue! Apparently sometimes the Zabbix server is still starting when the test tries to add the hosts, resulting in a failure. I added a 15 seconds wait, any suggestion for a cleaner/more elegant solution?

bastelfreak:
@d1nuc0m thanks for debugging! Do you know if the zabbix server has any option for us to check if it's started properly?

TheMeier:
Since the acceptance tests actually create hosts using the API I think the tests are sufficient. Maybe a check for port 10051 can be added in spec/acceptance/server_spec.rb
```
 describe port(10051) do
   it { should be_listening }
 end
 ```

TheMeier (approved):

d1nuc0m:
@TheMeier the problem is it listens after some time, is it possible to add a wait or something like that in the tests?

TheMeier:
I thought that is what this https://github.com/voxpupuli/puppet-zabbix/pull/1002/files#diff-b0943f1b9eb067ff42fd021d0b7f98cb415ffedc5df370441c5af0ed72b3de97R56 is supposed to do

PR #1026, opened 2026-02-20
feat: add Zabbix 7 API compatibility and proxy group support

mouchymouchy:
Update the zabbix_host provider to use the new Zabbix 7 API fields:
- Replace deprecated proxy_hostid with proxyid
- Add proxy_groupid and monitored_by fields for proxy group support
- Add proxygroup property to zabbix_host type
- Fix interfacedetails insync? comparison to use sorted comparison
- Add monitored_by_group parameter to agent class and resources

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1376,6 +1376,7 @@ The following parameters are available in the `zabbix::agent` class:
 * [`zabbix_package_source`](#-zabbix--agent--zabbix_package_source)
 * [`manage_resources`](#-zabbix--agent--manage_resources)
 * [`monitored_by_proxy`](#-zabbix--agent--monitored_by_proxy)
+* [`monitored_by_group`](#-zabbix--agent--monitored_by_group)
 * [`agent_use_ip`](#-zabbix--agent--agent_use_ip)
 * [`zbx_groups`](#-zabbix--agent--zbx_groups)
 * [`zbx_group_create`](#-zabbix--agent--zbx_group_create)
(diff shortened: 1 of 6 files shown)


cvquesty (approved):
I got a little lost in the toe, so I'm going to lean on the quality of your tests to tell me "it's ok".  Otherwise, all looks good.

teluq-pbrideau:
I tried your branch, but it failed compatibility with puppetapi `5.0.0-alpha1`.
Should you include the new `7.0.0-alpha4` version to this PR in the `zabbix::zabbixapi` class?  Or should it be in another PR maybe?

I did not find reference to this anywhere, but sorry if this is a duplicate I missed...

I just opened an issue for this version to be released on rubygems.org:
https://github.com/markt-de/zabbixapi/issues/129

teluq-pbrideau:
@mouchymouchy Sorry, I just found a bug in your PR.

When zabbix clients talk directly to the server, this error is raised:
```
 {
  "code": -32602,
  "message": "Invalid params.",
  "data": "Invalid parameter \"/1\": the parameter \"proxyid\" is missing."
}
```

You define the `monitored_by` as 1 or 2, but when there is no proxy, this should be set to 0
https://www.zabbix.com/documentation/7.0/en/manual/api/reference/host/object
> #### monitored_by
> Source that is used to monitor the host.
> Possible values:
> 0 - (default) Zabbix server;
> 1 - Proxy;
> 2 - Proxy group.


```
    monitored_by = proxy_groupid && proxyid.nil? ? 2 : 1
```

If I enter manually `monitored_by: 0` instead of your change in this PR, it works as expected.  Please fix it so `monitored_by` is 0 when there is no proxy configured.

EDIT: just to be clear, the problem is in this file: `lib/puppet/provider/zabbix_host/ruby.rb` on the new line 92.
Solution also could be to delete the entry, as 0 is the default value: `host_params.delete(:monitored_by) if host_params[:proxyid].nil?`.  Not necessarily the cleanest, but it also worked in my tests.

PR #1036, opened 2026-09-08
feat: manage PHP dependencies when necessary

### LINKED ISSUE #974: Frontend require different php version

- Puppet:
- Ruby:
- Distribution:
- Module version:

try to install frontend on el8 with zabbix 6.4 or 7.0

## What are you seeing
default php was 7.0

php version required by zabbix frontend

d1nuc0m:

Zabbix >= 6.0.47 requires PHP 7.3, default on EL8 is PHP 7.2

Reference: https://www.zabbix.com/rn/rn6.0.47

Manage it in a separate class that can be later removed (or kept for when it'll happen on EL9)

Fixes #974

### DIFF
--- a/data/os/RedHat/8.yaml
+++ b/data/os/RedHat/8.yaml
@@ -1,0 +1,4 @@
+---
+# https://www.zabbix.com/documentation/6.0/en/manual/installation/upgrade_notes_6045
+# "(...) requires PHP 7.3 or later, regardless of whether you are using SAML authentication."
+zabbix::manage_php: '8.2'
--- a/manifests/php.pp
+++ b/manifests/php.pp
@@ -1,0 +1,27 @@
+# @summary
+#   Install a version of PHP supported by Zabbix on system where the default one is too old
+# @param manage_php
+#   If set, version of PHP to install on systems where default one is not supported by Zabbix (actually only EL8).
+#
+class zabbix::php (
+  Optional[String] $manage_php = undef
+) {
+  if $manage_php {
+    case $facts['os']['family'] {
+      'RedHat': {
+        case $facts['os']['release']['major'] {
+          '8': {
+            package { 'php':
+              ensure   => $manage_php,
+              provider => 'dnfmodule',
+            }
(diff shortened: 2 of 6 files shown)


manifests/php.pp
```diff
@@ -0,0 +1,21 @@
+# @summary
+#   Install a version of PHP supported by Zabbix on system where the default one is too old
+# @param manage_php
+#   If set, version of PHP to install on systems where default one is not supported (actually only EL8).
```
evgeni: The code below doesn't restrict it to EL8 tho.
And on EL10 this will be an error if the users tries to set a version, as there are no more modules in EL10 (god bless).

evgeni: I'd probably restrict the code to only work on EL8 and EL9 and adjust the doc string to tell exactly that

data/os/RedHat/8.yaml
```diff
@@ -0,0 +1,4 @@
+---
+# https://www.zabbix.com/documentation/6.0/en/manual/installation/upgrade_notes_6045
+# "(...) requires PHP 7.3 or later, regardless of whether you are using SAML authentication."
+zabbix::manage_php: '7.3'
```
evgeni: According to https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle#rhel8_application_streams 7.3 is EOL since 2021, can Zabbix run with  8.2 which is supported until 2029 in RHEL and friends?

evgeni: or 7.4? that also has "full lifecycle" support

evgeni: image

manifests/php.pp
d1nuc0m: Should be limited now

data/os/RedHat/8.yaml
d1nuc0m: Theoretically, yes - let's check if tests do pass

evgeni (approved):

PR #1038, opened 2026-09-11
refactor: create test matrix with GH metadata

### LINKED ISSUE #937: Refactoring test

- Puppet: 7, 8
- Ruby: 2.7, 3.x
- Distribution: Archlinux and all other
- Module version: 10.0.0

Run acceptance on one of the operating system

## What are you seeing
All Oses run a test with default version and N for the supported version
In most case it run twice with the some version, in other case like Archlinux run only with operating system repo version of package that not in the supported list.

Run acceptance test only one per version
Run acceptance only on supported version

## Output log
https://archlinux.org/packages/extra/x86_64/zabbix-agent/
zabbix-agent 6.4.12-2

Need a plan to use in the best way acceptance test, Probabli custom matrix not solve the problem, but run test on Archlinux with parameter zabbix_version 5 and 6 with the installed package as 6.4 isn't the best we can do

d1nuc0m:

Drop custom matrix iterator, use pipeline tools

Fixes #937


d1nuc0m:
~~I'm not sure of how to skip tests on Arch~~

spec/acceptance/agent_spec.rb
```diff
@@ -2,169 +2,169 @@
...
+    end
   end
+end
 
-  describe "zabbix::agent class with agent2 and zabbix_version #{version}" do
-    # <6.0 agent2 packages are not available for ubuntu 22.04
-    next if version < '6.0' && default[:platform] =~ %r{ubuntu-22}
+describe "zabbix::agent class with agent2 and zabbix_version #{zabbix_version}" do
+  # <6.0 agent2 packages are not available for ubuntu 22.04
+  next if zabbix_version < '6.0' && default[:platform] =~ %r{ubuntu-22}
```
bastelfreak: we only test on zabbix 6 and 7, so this condition can be removed?

spec/acceptance/zabbix_template_spec.rb
```diff
@@ -3,66 +3,68 @@
 require 'spec_helper_acceptance'
 require 'serverspec_type_zabbixapi'
 
+zabbix_version = ENV.fetch('BEAKER_FACTER_zabbix_version', '7.0')
+
 describe 'zabbix_template type' do
```
bastelfreak: it would be a bit cleaner to use a condition on the describe, like:

```suggestion
describe 'zabbix_template type', unless:  zabbix_version >= '7.0' do
```

then the tests wouldn't even be in the suite.

spec/acceptance/agent_spec.rb
d1nuc0m: Yes, there are a lots of EOL-related leftovers, caught some other ones. Tests still need a lot of review though, some have hardcoded Zabbix versions, and/or don't support new versions...

spec/acceptance/zabbix_host_spec.rb
```diff
@@ -3,339 +3,341 @@
...
-          zabbix_api_user  => 'Admin',
-          zabbix_api_pass  => 'zabbix',
-          apache_use_ssl   => false,
-          manage_resources => true,
-        }
-      EOS
+  # Zabbix 7.0 removed the deprecated params 'user' in favor to 'username'
+  next if zabbix_version >= '7.0'
+  # Requires zabbix::web, that is unsupported on Arch
+  next if default[:platform] =~ %r{archlinux}
```
bastelfreak: could you move those conditions to the `describe` in line 9?

spec/acceptance/zabbix_hostgroup_spec.rb
```diff
@@ -3,65 +3,66 @@
...
-          zabbix_api_user  => 'Admin',
-          zabbix_api_pass  => 'zabbix',
-          apache_use_ssl   => false,
-          manage_resources => true,
-        }
-      EOS
+# Zabbix 7.0 removed the deprecated params 'user' in favor to 'username'
+describe 'zabbix_hostgroup type', unless: zabbix_version >= '7.0' do
+  # Requires zabbix::web, that is unsupported on Arch
+  next if default[:platform] =~ %r{archlinux}
```
bastelfreak: can you move that also to the `describe`?

spec/acceptance/zabbix_proxy_spec.rb
```diff
@@ -3,190 +3,191 @@
...
-        it 'has a interfaces port configured' do
-          expect(proxy2['interface']['port']).to eq('10055')
-        end
-
-        it 'has a interfaces useip configured' do
-          expect(proxy2['interface']['useip']).to eq('0')
-        end
+describe 'zabbix_proxy type', unless: zabbix_version >= '7.0' do
+  # Requires zabbix::web, that is unsupported on Arch
+  next if default[:platform] =~ %r{archlinux}
```
bastelfreak: same here

d1nuc0m:
Now passes, let's see removing hardcoded zabbix versions

d1nuc0m:
@bastelfreak here we are - I hate hardcoded values :D

marcusdots (approved):
LGTM.

